PR #3, opened 2019-12-30
Crude UI sounds

GraionDilach:
This adds sounds to the menu at point. Source included. I made this on my own. Feel free to apply which license you intend to apply.


Mailaender:
Thanks. I suggest https://creativecommons.org/licenses/by/3.0/ to not complicate the matter and you may want to add yourself to the AUTHORS file.

GraionDilach:
CC-BY sounds cool because there are a lot of free assets for it on Newgrounds & co. and has no explicit requirement at providing sources (unlike a GPL license).

However, since these particular ones were relatively simple to do, I feel there's no need to require credit for these and set up the folder structure to indicate their CC0 license (which requires no credit).

PR #8, opened 2020-01-04
Radar Building

Dzierzan:
A building taken from Tyrian, remapped and added custom damage state.


Mailaender:
Thanks. You may want to add yourself to the `CREDITS` file.

PR #14, opened 2020-01-10
Added factory

Dzierzan:


Mailaender:
Added in https://github.com/Mailaender/OpenHV/commit/247b07e3b9319b4102d724d36db9a56e841989b8 with the necessary metadata for the frames.

This remap was badly needed as the original dark blue parts used in the animations don't remap well.

PR #15, opened 2020-01-10
Added airport pngs

Dzierzan:


Mailaender:
I prefer to use the original file names although we may use it for something else.

Mailaender:
Had to re-arrange the frames, fix a pixel error and change the yaml files. https://github.com/Mailaender/OpenHV/commit/0779f265e501d9a70c833a30f4f72fb9bfff8917

PR #17, opened 2020-01-10
Add 2nd remapped factory building

Dzierzan:


Mailaender:
Added in https://github.com/Mailaender/OpenHV/commit/23030f487bced373d286b5f02e50d9b852b80a65. You forgot about the utilty command.

PR #19, opened 2020-01-10
Remapped Generator

Dzierzan:


Dzierzan:
Yeah, I just made it more "compact". The FramSize has to be updated which is 39,52. I should have updated yaml file as well.

Mailaender:
639 / 40 = 15,975 px so that won't, too.

Dzierzan:
Check the generator now.

Mailaender:
https://github.com/Mailaender/OpenHV/commit/e6e44231d82569694a448f663fd547d0946dcbf2

PR #20, opened 2020-01-10
Add remapped turret png

Dzierzan:


Mailaender:
Added in https://github.com/Mailaender/OpenHV/commit/e00ebc84ad022955dfb896d5ce11ce0a62170cdd.

PR #22, opened 2020-01-10
Remapped flag

Dzierzan:


Dzierzan:
But you can update FrameSize (39,52) in yaml file which should fix the problem

Mailaender:
Your flag is W: 84px | H: 20px not 80x20.

Dzierzan:
Sorry, you're right. It should be 17x20 per frame.

Mailaender:
That would be 85px then, but the file is only 84px.

GraionDilach:
Related question: why the shadows were dropped from the flag anyway? The original sprite in HV included shadows.

Mailaender:
Not all units have shadows and the offsets don't look quite right on first sight so I skipped them for now. Might add them later.

Mailaender:
https://github.com/Mailaender/OpenHV/commit/fb39752ae80a3a6c85dfd941363930ee9fd1bfdc I kept it 20x20 px as my existing offset calculations depend on it.

PR #23, opened 2020-01-10
Add remapped Module building

Dzierzan:


Mailaender:
Again and odd choice of width and height. Remember this is a 20x20 tiles game. This is what I did:

image

I use a grid/raster line of 40x50 (2x2 building with some height) in GIMP to check of I am not off by a few pixels.

Dzierzan:
From my point of view, it's all about having the smallest rendering dimentions, while imperfections can be adjusted with Offset parameter.

Mailaender:
https://github.com/Mailaender/OpenHV/commit/4b84b4100b218ecd4e62536b5db4f5511300f02c

PR #24, opened 2020-01-10
Add remapped Miner unit

Dzierzan:


Mailaender:
I chose 20x20 which seems more sensible and what the artist intended it to be. https://github.com/Mailaender/OpenHV/commit/85171fb283d3629ed5a5fbc4b349078fb19415ed

Issue #4, opened 2019-12-31
Clarify attribution with an explicit futureproof crediting guideline/standard to satisfy CC-BY requirements with an OS mindset.


GraionDilach:
The CC-BY license is a good choice for the assets for this project. And as long as there is only a single artist whose assets are involved, a generic crediting message is enough.

The problem comes in if multiple additional artists starts contributing and/or people would start derivative works of the assets within the project, for example recoloring/remixing tiles to extend the tileset. There should be a way to effectively maintain who was the artist of a particular asset but maintaining this shouldn't happen within the ingame credits list because that would overload the widget.

A suggestion I can think of is to have a readme.md in every subfolder within bits where actual assets are which holds this information of ownership for each file (either separately, either with wildcards) and then demand outside asset contributions to include this information within their PR. This then allows for outside derived works to maintain the attribution required by the license and does not force them to fall back to a generic crediting and/or resulting with unclear or confusing licensing while also contributors can be ensured that their work does not get misattributed in the future either.

I know I can be a bit of annoyance with such an issue like above in the project's early state and i don't want to sound demanding so feel free rejecting it if you're thinking otherwise.

Mailaender:
We may be able to auto-generate this from git where it is self-documented already.

```bash
git log --name-status --diff-filter=A --format='> %aN' \
| awk '/^>/ {tagline=$0}
       /^A\t/ {print tagline "\t" $0}'
```

PR #30, opened 2020-01-12
Add some CC 3.0 license sounds

Dzierzan:
Added some sounds from soundbible website. All authors are credited in README.md


GraionDilach:
Could you please change the folder setup? Spaces in the path are probably a bad idea.  #4 provides some ideas.

GraionDilach:
Sorry, but when I meant spaces in path, I also meant spaces in file names.

GraionDilach:
I'll try to go through these during tonight. Might file a PR against your PR - I have filed all sounds to this repository with a shared volume setting so far (GoldWave's Normalize Volume to -16 dB) which sounds like an idea to do here as well.

I probably should document those above settings anyway.

GraionDilach:
These sound way more professional than what I'd expect being released as CC-BY. Tomorrow I'll recheck their sources on that SoundBible one-by-one. Sorry, but these sound suspicious on first hear.

They are also huge anyway and need to be downsampled. I went with 22kHz IMA mono in the other PR because variants of that format was very common back during 1995-1999.

Dzierzan:
I edited them too by cutting too long parts or speeding up the tempo (missile one). You can check if their license is correct. Besides classic graphics do not mean you are forced to use 22kHz.

GraionDilach:
> do not mean you are forced to use 22kHz

https://github.com/OpenRA/OpenRA/issues/14153 is still an issue and it can either be based upon the amount of channels utilized or the amount of data required to be puffered. Since the exact cause of it is still unknown, it's better to play safe and prevent wasting 800 kB in the sound puffer for a single sound effect if you can decrease that to ~20 kB first.

Additionally, sound quality should remain consistent among all the samples used, considering the visuals seem to achieve a fairly consistent style so far.

I don't see any reason why not to go with 22 kHz mono as the standard. The entire pre-C&C3 C&C modding community has no issue with that format, even when this "degrades" voice actors who ventured to/came from non-union professional works (Gabriel Wolf, who was a prominent voice actor at Star Trek Online before it went to union voiced a good portion of Mental Omega 3.0, not to mention George "Seóras" Exley and Ki "ComradeCrimson" McKenzie whom kickstarted their career from C&C/Warhammer 40k mod roles, notably Rise of the Reds). I will not deny that there is indeed a minor fidelity loss with the downsampling, but I don't think that difference can be audible ingame.

Dzierzan:
I didn't know such details though. If it's a must to use 22kHz, then so be it.

GraionDilach:
Superseded by #42.

Mailaender:
Not a fan of adding all those unused sounds, but okay.

PR #31, opened 2020-01-12
Include sounds sampled from the qubodup military pack.

GraionDilach:
Original source: https://freesound.org/people/qubodup/packs/4366/

Note: I intentionally added more sounds than what the project needs currently. This allows room for improvements among the available options if you disagree with my choices.

There might be a followup of this where I go over and include additional ones from this pack but for now this sounds enough. The radio beep might be useful for EVA text alerts along in a followup.


Mailaender:
These fit in pretty well. Good findings.

PR #32, opened 2020-01-12
Added a skirmish testmap.

GraionDilach:


Mailaender:
I also have a test map https://www.dropbox.com/s/1ke3oa2pbhoxseo/test.oramap?dl=0 which is a bit more advanced as I test the terrain in it. Due to https://github.com/Mailaender/OpenHV/issues/26 and the tileset being WIP I don't accept maps yet. I suggest you save yours in the user folder for now.

Cherry picked the player radar fix. Thanks!

GraionDilach:
Ah, okay. I'm using this map locally to check out the actors anyway.

PR #33, opened 2020-01-12
Remapped Scout1

Dzierzan:
Remapped scout unit. Works in game.


Mailaender:
This looks a tiny bit better. =)

PR #36, opened 2020-01-17
Remapped Tank5

Dzierzan:
This is just Tank5d.bmp remap with minor changes.


Mailaender:
You are changing the boundaries:

```diff
-W: 160px | H: 20px | Size: 2.22 KB
+W: 168px | H: 21px | Size: 4.72 KB
```

Dzierzan:
Because of West-South direction is 21 px wide. You can check the original one to see that.

Mailaender:
Okay, I was irritated because you didn't update the `tank5.yaml`.

Dzierzan:
Right. I keep forgetting about those.

Mailaender:
https://github.com/Mailaender/OpenHV/commit/85fc618c2fdfe80044ff80a32e3675b41c461e8c It looks way better with the partial remap.

PR #42, opened 2020-01-19
Add a CC-BY-3.0 audio folder and various sounds proposed by Dzierzan.

GraionDilach:
Supersedes #30. I went ahead and reorganized the file structure for future maintainability, also included source link in each commit for reference, to follow my principle in #31.

Besides audio normalization and downsampling, there are a few additional differences.
 - The heavy gun fire sound's fadeout was sped up, so the overall length of the sound is now 1.5 seconds instead of the proposed 3 seconds.
 - The grenade launching sound's left channel was dropped because it started with a popping for some reason. Since this didn't affect the right channel, I used that one as the source for the mono channel.


Mailaender:
Merged this manually.

PR #43, opened 2020-01-19
Use the Terminal Grotesque pixel font for ingame texts.

GraionDilach:
Source: https://github.com/raphaelbastide/Terminal-Grotesque

image


Mailaender:
I like the idea of having a custom pixel art font, but it seems to expose some rendering issues:

image

PR #44, opened 2020-01-19
Misc. fixups

GraionDilach:
Some misc stuff I did today.

### DIFF
--- a/mods/hv/rules/buildings.yaml
+++ b/mods/hv/rules/buildings.yaml
@@ -212,7 +212,7 @@ MODULE:
 
 MINER2:
 	Inherits: ^BaseBuilding
-	Inherits@shape: ^2x2Shape
+	Inherits@shape: ^2x3Shape
 	Selectable:
 		Bounds: 40,60
 	Tooltip:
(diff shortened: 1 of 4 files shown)


mods/hv/rules/buildings.yaml
```diff
@@ -226,9 +226,11 @@ MINER2:
 		Sequence: active
 	UpdatesDerrickCount:
 	Building:
+		Footprint: xx xx xx
+		Dimensions: 2,3
 		RequiresBaseProvider: false
-		TerrainTypes: Ore
-		AllowInvalidPlacement: true # TODO allow placement only on ressources
+		TerrainTypes: Clear, Road, Crater, Grass, Ore
+		AllowInvalidPlacement: true # TODO CanPlaceBuilding explicitly prevents placing upon tiles with resources
```
Mailaender: Yeah I know. I plan to patch that.

mods/hv/rules/buildings.yaml
```diff
@@ -226,9 +226,11 @@ MINER2:
 		Sequence: active
 	UpdatesDerrickCount:
 	Building:
+		Footprint: xx xx xx
+		Dimensions: 2,3
 		RequiresBaseProvider: false
-		TerrainTypes: Ore
-		AllowInvalidPlacement: true # TODO allow placement only on ressources
+		TerrainTypes: Clear, Road, Crater, Grass, Ore
```
Mailaender: Can you please revert that?

GraionDilach: Why? Won't this be required for longterm?

I thought the idea is to allow placing the miner building upon single-cell patches of resource as well - in which case the other cells will need to remain plain buildable cells to fit the extractor in and those cells will also need to satisfy the building's own validity check.

The GrantConditionOnTerrain check prevents the miner already to deploy to positions where no resource is under it, this just gives more room for building placement (assuming the CanPlaceBuilding issue is fixed).

Mailaender: You are right, the miner is not 1x1.

mods/hv/rules/buildings.yaml
Mailaender: #17838

PR #50, opened 2020-01-26
Add shadows for Buildings

Dzierzan:
unknown

Still misses Silo and Miner. I will add those later.


GraionDilach:
The Main Base seems to have a different shadow direction and probably should also be directed towards bottom right instead of top-right.

Dzierzan:
It is original one, yes, it is weird. I might resolve that in the future.

Mailaender:
Squashed and fixed this a bit https://github.com/Mailaender/OpenHV/commit/eac417fa72be2becb611fc063a03c263808ba801.

PR #52, opened 2020-01-28
Improved spritesheet of Factory2

Dzierzan:
factory2
It has now less green color.


Mailaender:
Good choice. Less is more here.

PR #58, opened 2020-01-31
Add crater smudges

Mailaender:
Closes https://github.com/Mailaender/OpenHV/pull/57.


GraionDilach:
Why did you cut it up?

Mailaender:
I looked that the Westwood games mods and thought that was required somehow.

GraionDilach:
We don't have to follow WW all the time afterall - I guess that's one of the actual aims even.

On another thought, this opens up possibilities to later extend the shorter smudges with additional frames which does not necessitate `Combine` nor entire sheet  reupload, unnecessarily bloating the repo's size. So yeah, you're better with this.

PR #60, opened 2020-02-01
More Custom sparks

Dzierzan:
Custom sparks plus minor changes for Scout. Made its projectile InstantHit instead of Bullet and added some inaccuracy.

new sparks


GraionDilach:
You coulda expand on this via having additional `FlipX` variants of all 4 within the sequence yaml, but that's just me.

PR #70, opened 2020-02-17
Remapped Boat2

Dzierzan:


mods/hv/mod.yaml
```diff
@@ -1,197 +1,197 @@
...
-	SupportedExtensions: .png
-
-GameSpeeds:
-	default:
-		Name: Normal
-		Timestep: 40
-		OrderLatency: 3
+Metadata:
+	Title: OpenHV
+	Version: f45dd24781cb56f4cef4a9e1ffe74b4b12f741be
```
Mailaender: I think you hit `make version` or something. Please don't commit this.

PR #74, opened 2020-02-23
Added smaller water splash effect for firearms.

Dzierzan:
small water splash


GraionDilach:
Is there an actual need for the second row? Looks like a flipped version of the first, and then having that first row with FlipX:true in sequences covers it.

Dzierzan:
It's flipped with minor changes.

PR #84, opened 2020-04-10
Remapped Silo

Dzierzan:
Added also lights and orb animation, and damaged state.


Mailaender:
The shadow contains all the buildings and is not set up.

mods/hv/sequences/buildings.yaml
```diff
@@ -176,8 +176,22 @@ field:
...
+		Start: 0
+	animation_lights: silo.png
+		Start: 2
+		Length: 4
+		Tick:	240
+		Reverses: true
+	animation_orb: silo.png
+		Start: 6
+		Length: 7
+		Tick:	60
```
Mailaender, line 188: You added a tab before setting the value which is odd.

mods/hv/sequences/buildings.yaml
```diff
@@ -176,8 +176,22 @@ field:
...
+		Reverses: true
+	damaged-idle: silo.png
+		Start: 1
 	icon: blank.png
 
+
```
Mailaender, line 194: Stray new line.

PR #88, opened 2020-04-27
A new voiceset

Dzierzan:
I tried my ugly voice and created a new voice set. Currently it works only for Vehicles.


GraionDilach:
Not bad, not bad at all. I'm genuinely impressed.

Should've done a soft noise removal before cutting up the master recording though, since there's no clear-cut decision on the voice direction yet. You might also want to prefix the files themselves incase you want to submit a set with a different accent at a later point. Voice quality is pretty okay otherwise and you didn't made the rookie popping mistakes. The lines are also normalized together. Great job for a first, pretty much the amount of noise is the sole issue I can hear.

My two generic sets were never meant to be final sets anyway, just placeholders which I expect being phased out when unit roles start getting defined. Calling this `GenericVoice03` might even be a mistake since this is a clear step forward from those and calling it `VehicleVoice01` to implicate it can be kept around properly might be better.

Dzierzan:
Thanks!

Noise is there because I added it. I wanted to make it more vintage, as voice heard through radio. I wanted to add that radio sound heard in Dune2000 as well, but I couldn't find anything good.

Yeah, there are no pops because I've got a makeshift popkiller.

GraionDilach:
Ah, so that's intentional. It's not a bad idea but I hope you asked @Mailaender about this beforehand, because there wasn't a reference set before to copy the voice quality/style from.

I'm fine with this being the reference, but then please note down the noise level for future contributors to ensure future consistency.

PR #94, opened 2020-05-04
New Copter

Dzierzan:
This PR changes the copter, so now the body is seperated from its propeller. Plus under the copter there's a rotatable turret.


mods/hv/rules/aircraft.yaml
```diff
@@ -62,20 +62,23 @@ COPTER:
...
-	AttackAircraft:
-		FacingTolerance: 20
-		PersistentTargeting: false
-		AttackType: Hover
-		OpportunityFire: true
+	Turreted:
+		TurnSpeed: 20
+		Offset: 200,0,-100
+	AttackTurreted:
+		PauseOnCondition: ecmdisabled
```
Mailaender, line 69: Remove that. Helicopters can't be disabled by the ECM tank.

PR #112, opened 2020-07-10
Tech center

Dzierzan:
Basic implementation. It does nothing at the moment.

techcenter


mods/hv/rules/buildings.yaml
```diff
@@ -509,6 +509,49 @@ FACTORY3:
...
+		HP: 30000
+	Armor:
+		Type: Steel
+	RevealsShroud:
+		Range: 4c0
+	Power:
+		Amount: -200
+	Targetable:
+		TargetTypes: Ground, Structure
+	ScalePowerWithHealth:
```
Mailaender, line 541: Only power plants should have this.

Mailaender:
Also needs a rebase.

PR #114, opened 2020-07-10
Seperated cursors, updated credits.

Dzierzan:


mods/hv/cursor.yaml
```diff
@@ -1,78 +1,78 @@
-# License: CC-BY-SA-4.0
+# License: CC-BY-SA-4.0, CC-BY-3.0-US
```
Mailaender, line 1: This is the cursor definition file. Not about the assets.

PR #122, opened 2020-07-19
Remove BOM marks.

GraionDilach:
This normalizes all C# source files to lack a BOM.


Mailaender:
Confirmed with `grep -Rl $'\xEF\xBB\xBF' .` and noticed that some files from the engine contain it.

PR #125, opened 2020-07-31
Miscellaneous fixups

GraionDilach:


mods/hv/weapons/missiles.yaml
```diff
@@ -149,7 +143,6 @@ ShipMissile:
 	ReloadDelay: 50
 	Range: 9c0
 	MinRange: 2c0
-	Report: missile7.aud
```
GraionDilach, line 152: I've seen that `rocketlaunch01` is the generic choice so I let it falling back to that one here.

Mailaender:
Thanks.

PR #126, opened 2020-08-02
Building damagefires.

GraionDilach:
This is a bit of a draft at my side, because I'm not certain which model should be followed here, C&C's explicit burning locations, of Warcraft 1/Warcraft 2 where the fire was burning in the center of the building artwork. If we agree on the former, I'll set up offsets.

Additional point of discussion would be the art quality, since I decided to squash the flame's height. Pinging @Dzierzan for thoughts on this matter.

I'm also planning to do sparking for vehicles in a separate PR.


Mailaender:
I think this needs to be tweaked at a per building level as it is too uniform and the positioning is sometimes a bit out of place:
image

Mailaender:
Then the C&C system may be better. The Warcraft one only really works for the forest fires.

GraionDilach:
Updated.

hv-2020-08-08T140354692Z

mods/hv/sequences/buildings.yaml
```diff
@@ -1,6 +1,30 @@
 # License: CC-BY-SA-4.0
 
+^BuildingFlame:
+	flame: flame2.png
+		Length: *
+		ZOffset: 512
+		Offset: 0, -9
```
Mailaender, line 7: Why didn't you use the `WithShadowOverlay` `Offset` instead?

sources/artwork/Flame2
```diff
@@ -0,0 +1,6 @@
+Flame.lbm
+last coloumn cropped
+Width: resized to 50 pixel
+Height: resized to 13 pixel
+fourth frame dropped
+minor misc. pixel tweaks
```
Mailaender, line 6: Keeping a changelog for each image editor action seems a bit over the top.

mods/hv/sequences/buildings.yaml
GraionDilach: I was unaware of that option, but even relying on that would be wrong. These are art decorations, they shouldn't be offset in world units, but in pixels directly. Converting to world units back and forth to get proper values sounds wrong to me (these values were mostly calculated with Photoshop's Ruler tool on the artwork directly).

GraionDilach, line 7: This default value is to ensure the flame will be played in the cell's center position. If this should be mentioned in the sequence itself I'll update.

sources/artwork/Flame2
GraionDilach: Thought I should document everything.

mods/hv/sequences/buildings.yaml
Mailaender: Mentioning it here is enough documentation for me.

PR #128, opened 2020-08-02
Add condition support to WithConstructionBeamOverlay.

GraionDilach:


Mailaender:
Thanks. Seems I forgot to push that change. I also didn't use invalid token properly in my variant.

GraionDilach:
You're welcome then. 😄

OpenRA.Mods.HV/Traits/Render/WithConstructionBeamOverlay.cs
```diff
@@ -58,9 +64,20 @@ public WithConstructionBeamOverlay(ActorInitializer init, WithConstructionBeamOv
...
+				token = self.GrantCondition(info.Condition);
+
 			// Remove the animation once it is complete
 			if (overlay != null && renderSprites != null && anim != null)
-				overlay.PlayThen(info.Sequence, () => self.World.AddFrameEndTask(w => renderSprites.Remove(anim)));
+			{
+				overlay.PlayThen(info.Sequence, () => self.World.AddFrameEndTask(w =>
+				{
+					if (token != Actor.InvalidConditionToken)
+						token = self.RevokeCondition(token);
```
Mailaender, line 76: Actually this is wrong. I had the reverse in mind.

Mailaender: Yours makes more sense though.

PR #129, opened 2020-08-02
Set all missile CruiseAltitude to 2c0.

GraionDilach:
This prevents them flying higher than the actual aircraft,
causing a weird path when they are catching up with their target.


Mailaender:
This was actually some kind of on purpose as the tank shoots upwards. That effect is still there just not so over the top. Guess that's better then.

PR #133, opened 2020-08-08
Add a new drone voiceset for the pods.

GraionDilach:
Technically this is a two-years old set which I recorded for a drone in AS-as-a-mod, but since that AS is dead, I thought this might fit OpenHV.


Mailaender:
That sounds pretty alien. I actually thought of scout pods as one man vehicles.

GraionDilach:
I thought they're supposed to be robots. Okay, nevermind then.

mods/hv/audio/voices.yaml
```diff
@@ -25,3 +25,9 @@ GenericVoice3:
 		Select: awaiting_orders, im_ready, receiving, yes_sir
 		Move: acknowledged, affirmative, as_you_wish, at_once, en route, lets_go, move_out, on_my_way, orders received, ten_four
 		Action: attack, destroy_them, hit_them_hard, light_em_up, target_locked, you_got_it
+
+PodVoice:
```
Mailaender, line 29: Let's use them for the drone ship instead.

Dzierzan:
Sounds like an angry robot dog.

mods/hv/audio/voices.yaml
GraionDilach: Alright, I've already rebased this branch locally, just let me find the actor.

PR #134, opened 2020-08-08
Add a CC0 sound for the nuke explosion.

GraionDilach:
Tweaked variant of https://freesound.org/people/tommccann/sounds/235968


Mailaender:
That indeed sounds scarier. Thanks.

PR #135, opened 2020-08-08
Replace the repair wrench with a ratchet.

GraionDilach:


Mailaender:
The artwork looks way better, but contrast seems to be a problem.

GraionDilach:
Increased the contrast on the handle. The head seemed to be fine from the get-go.

Mailaender:
Maybe add a white outline? The grey looks very similar to the building grey.

GraionDilach:
Everything has dark outlines in the source artwork. I'm not sure how to proceed.

GraionDilach:
Another option for a followup could be to recolor the handle into player color. OpenRA has ally-repair and I'm not certain if it's enabled here (or should be).

Mailaender:
That is a good idea. We probably want ally repair. I just didn't have art.

PR #139, opened 2020-08-09
Add sparking to damaged units.

GraionDilach:
Artwork also has an unused electric sparks for now. I'll file a followup at a point to add warheads to utilize it.


Mailaender:
They look great. Nice attention to detail.

PR #140, opened 2020-08-09
Hitshape fixups

### LINKED ISSUE #132: Various inconsistencies between structure hitboxes, artwork size and targeting positions
Something I noticed during setting up #126. There are multiple issues here.

Personally I'd propose that hitboxes should be aligned to cells and targeting offsets should be the center of cells with buildings. This is a step backwards compared to what OpenRA allows, but this would also ensure that customized offsets wouldn't provide edge cases in gameplay balance.

image

GraionDilach:
Still work in progress.

This PR's aim to normalize footprints, selection boxes, targeting offsets and hitshapes **to cover and align to the cells the building actually covers on the ground**. If necessary, I can incorporate the cells the building might cover due to it's height (Nuke Command Center and Miner are the most notable cases here) into footprints and selection boxes only.  

Fixes #132.


mods/hv/rules/buildings.yaml
```diff
@@ -2,18 +2,17 @@
...
 	Valued:
 		Cost: 3500
 	Tooltip:
 		Name: Main Base
 	Building:
-		Footprint: xxx xxx xxx
-		Dimensions: 3,3
-		LocalCenterOffset: 0,-512,0
+		Footprint: xx xx
+		Dimensions: 2,2
```
Mailaender, line 15: This might need a ZOffset correction in the sequence as it now collides with other buildings:
image

It may also make sense to have the right side to be a traversable space as we lack bibs and the AI has trouble when cluttering it's base with buildings.

GraionDilach: Okay, I'll incorporate the cells being hidden as bibs.

GraionDilach:
This exposed the various issues caused by `WithIdleOverlay`->`Offset` being used for artwork pixel offsets. This will be a long project and might take a while to be completed.

GraionDilach:
Ready for review.

Mailaender:
Looks like you touched everything but the trade platform.

GraionDilach:
Yes, IIRC the bug with repair buildings having docking issues on even footprints is still there. In that case, it's just better to leave it as-is and fix it after that bug gets fixed.

PR #142, opened 2020-08-09
Add two AS warheads with usecases.

GraionDilach:


OpenRA.Mods.HV/Warheads/FireRadiusWarhead.cs
```diff
@@ -0,0 +1,109 @@
...
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Linq;
+using OpenRA.GameRules;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.VH.Warheads
```
Mailaender, line 17: `VH?`

OpenRA.Mods.HV/Warheads/WarheadVH.cs
```diff
@@ -0,0 +1,85 @@
...
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Mods.Common.Warheads;
+using OpenRA.Primitives;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.VH.Warheads
+{
+	[Desc("AS warhead extension class." +
+		"These warheads check for the Air TargetType when detonated inair!")]
+	public abstract class WarheadVH : Warhead
```
Mailaender, line 23: Call it `AirTargetWarhead` then maybe? Is this upstreamable?

OpenRA.Mods.HV/Warheads/FireRadiusWarhead.cs
GraionDilach: Fixed.

OpenRA.Mods.HV/Warheads/WarheadVH.cs
GraionDilach: It isn't, because it conflicts with https://github.com/OpenRA/OpenRA/pull/18072.

Updated with the functions from that proposal though, so if it's followup gets in, this file can be deleted outright.

Mailaender: I posted a followup at https://github.com/OpenRA/OpenRA/pull/18777.

PR #145, opened 2020-08-15
Added selection decorations for top, bottom, left and right 

Mailaender:
image
This depends on https://github.com/OpenRA/OpenRA/pull/18521.


GraionDilach:
Can there be a fixup commit here for the bottom left and right corner offsets to stop being misaligned compared to the actual selection/hitbox?

image

Mailaender:
Fixed it.

PR #146, opened 2020-08-17
Expand powerdown with allowing it to disable prerequisites.

GraionDilach:
This is an experimental change on my side partially to diverge from C&C (which never did this) and partially to strengthen/justify powerdown into an actual feature instead of being a gimmick.

Actual factories (Creation Platform/Harbor/Starport) are untouched for now but a followup could disable their production traits to allow them being powered down as well.


Mailaender:
Tiberium Wars did this and I am a huge fan of the idea.

PR #147, opened 2020-08-18
Remove UseClassicPerspectiveFudge from BodyOrientation.

GraionDilach:
https://github.com/OpenRA/OpenRA/blob/56739f87fb0ecf6df08e3e24367ddca4a896fd5e/OpenRA.Mods.Common/Traits/BodyOrientation.cs#L26 doesn't apply to us because we're using an actual topdown perspective (this is coming from RA1).

This breaks weapon firing offsets though which I haven't fixed yet.


Mailaender:
Good catch. I didn't know that defaults to true.

PR #149, opened 2020-08-18
Remove the tooltip from Mount Brosnan's lavalight.

GraionDilach:
This fixes the custom rules warning in the map selection.


Mailaender:
Removing the `Tooltip` which triggers the warning is enough. I must have forgotten to push that change.

GraionDilach:
Hm, true that. I thought the additional actor itself can already cause problems but apparently that's not the case, cool.

Mailaender:
To be fair I didn't come up with that. @pchote helped.

PR #154, opened 2020-08-21
Added new radar building

Dzierzan:
This new radar building is supposed to be used for 2nd faction, doesn't matter if it's YI or SC. It is based on Tyrian's building.

I haven't set up any 2nd faction yaml code and I don't think I will. So this PR simply adds new building to the pool.

dome2


Mailaender:
Can you put the original file in `sources/artwork/Tyrian/`?

PR #168, opened 2020-08-23
Add the pyramid from Tyrian.

GraionDilach:
image


mods/hv/rules/props.yaml
```diff
@@ -0,0 +1,55 @@
...
+	Health:
+		HP: 120000
+	WithIdleOverlay@shadow:
+		Sequence: shadow-overlay
+		Palette: shadow
+		IsDecoration: true
+
+PYRAMID2:
+	Inherits: ^Prop
+	Inherits@shape: ^2x2Shape
```
Mailaender, line 46:
```suggestion
	Inherits@shape: ^1x1Shape
```

GraionDilach: Whoops, fixed.

PR #170, opened 2020-08-23
Add new map Tomb Raiders.

GraionDilach:
image

Depends on #168. Only the last two commits are new.


GraionDilach:
Whoops, I have an error here. The small pyramid formulation on the middle right side is accidentally flipped.

GraionDilach:
Fixed. I won't update the map preview though from the first post - the left edge was adjusted to mirror the right edge.

GraionDilach:
Rebased.

Mailaender:
The gold in the middle is unreachable, because you walled it off with cliffs.

GraionDilach:
That is intentional and not unreachable. You're supposed to get a flying transport and a builder to place a storage there and then deliver miners manually.

PR #183, opened 2020-09-10
Set up the tech building into a Tech Reconstructor.

GraionDilach:
This is a bit more advanced than the RA1 setup, allowing for stacking building effects up to 5 instances, inspired by YR Hospital and Machine Shop.

AI setup is intentionally missing for now because I think I'll replace the AI capture module.


mods/hv/rules/buildings.yaml
```diff
@@ -1410,7 +1410,7 @@ TECHBUILDING:
 	Selectable:
 		Bounds: 40,48,0,-4
 	Tooltip:
-		Name: Tech Building
+		Name: Tech Rejuvenator
```
Mailaender, line 1413: Rejuvenator is a bit odd. How are tanks rejuvenated. Maybe Titanium Armor Research Facility or something?

GraionDilach: That suggestion sounds convoluted - all our other names are simple. Reconstructor is my other idea.

Mailaender: Yes, that's better.

GraionDilach: Updated with that proposal.

GraionDilach:
Fixed the accidental omission of sequence renaming which broke the build.

PR #186, opened 2020-09-19
Refactor AI capturing.

GraionDilach:


OpenRA.Mods.HV/Traits/BotModules/CaptureManagerHVBotModule.cs
```diff
@@ -0,0 +1,245 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using OpenRA.Mods.Common;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.HV.Traits
+{
+	[Desc("Manages AI capturing logic.")]
+	public class CaptureManagerHVBotModuleInfo : ConditionalTraitInfo
```
Mailaender, line 22: Maybe
```suggestion
	public class PriorityCaptureManagerBotModuleInfo : ConditionalTraitInfo
```

GraionDilach:
Applied the suggested name.

OpenRA.Mods.HV/Traits/BotModules/PriorityCaptureManagerBotModule.cs
```diff
@@ -0,0 +1,245 @@
...
+
+							capturers = capturers.Skip(1);
+						}
+
+						priorityTargets = priorityTargets.Skip(1);
+					}
+				}
+
+				if (!capturers.Any())
+					return;
```
Mailaender, line 184: This is duplicated and probably wrong here.

OpenRA.Mods.HV/Traits/BotModules/PriorityCaptureManagerBotModule.cs
```diff
@@ -0,0 +1,245 @@
...
+			}
+
+			var randPlayer = world.Players.Where(p => !p.Spectating
```
Mailaender, line 187:
```suggestion
			var randomPlayer = world.Players.Where(p => !p.Spectating
```

OpenRA.Mods.HV/Traits/BotModules/PriorityCaptureManagerBotModule.cs
GraionDilach: It goes to a line above, true that. This is an early return incase all capturers were used up for priority.

OpenRA.Mods.HV/Traits/BotModules/PriorityCaptureManagerBotModule.cs
```diff
@@ -0,0 +1,245 @@
...
+						var priorityTarget = priorityTargets.First();
+
+						var captureManager = priorityTarget.TraitOrDefault<CaptureManager>();
+						if (captureManager != null && captureManager.CanBeTargetedBy(priorityTarget, capturer.Actor, capturer.Trait))
+						{
+							bot.QueueOrder(new Order("CaptureActor", capturer.Actor, Target.FromActor(priorityTarget), true));
+							AIUtils.BotDebug("AI ({0}): Ordered {1} {2} to capture {3} {4} in priority mode.",
+								player.ClientIndex, capturer.Actor, capturer.Actor.ActorID, priorityTarget, priorityTarget.ActorID);
+
+							capturers = capturers.Skip(1);
```
GraionDilach, line 176: This should skip even if this block fails.

PR #189, opened 2020-09-19
Further capture polishes.

GraionDilach:
The second commit was added because queueing up to 15 targets stressed the pathfinder and led to lag spikes.


GraionDilach:
Dropped the third commit.

Issue #184, opened 2020-09-11
Project proposal: recolor the blue light animations to red for Synaty Corp.


GraionDilach:
The Spaceport sticks out currently, being the sole red lighted building among all of them. Having this change would help on visually differing the two faction from each other.

Dzierzan:
So far progress:

![image](https://user-images.githubusercontent.com/17529329/142498294-f5dac3b9-b327-4c67-842e-51cb28fa1138.png)

GraionDilach:
Think this is mostly done by now.

Dzierzan:
Only SC Tech center is remaining. Generator was updated in the meanwhile:

![generator2](https://user-images.githubusercontent.com/17529329/143581731-520c05a6-f3d3-418f-a12c-996dc299d36c.gif)

PR #193, opened 2020-09-27
Allow artwork reused with BGRA palettes for Synaty Corporation.

GraionDilach:
Proposal related to #184.

The current state of the PR is just a proposal for a direction.

image


GraionDilach:
The Starport is an interesting case here. During this BGRA proposal, I noticed that it has a slight red details on the building by default, so it probably needs some switcheroo to match the shades with the factions properly.

GraionDilach:
Current state:
image

Dropped the icon recolors - I still kept a local branch which has that separated out incase it would be given another try.

Mailaender:
This looks great and is really in line with DOS game floppy disk restraints and the choice of tricks used back in the days. For us, it means better maintainability. 👍🏻

PR #212, opened 2020-10-03
Added new splashes for Swamp and Lava, and updated valid targets.

### LINKED ISSUE #201: Lava cannot be targeted
Title^

Dzierzan:
As title says.

Fixes #201


Mailaender:
Filename `*splash.png.yaml` should be just `*splash.yaml`.

Mailaender:
You need to change the target type to swamp for this to work: https://github.com/OpenHV/OpenHV/blob/1c804edfd528671d6c337c5ed3e6d37ff3945cf6/mods/hv/tileset.yaml#L77-L80

Mailaender:
Now boats can't travel on swamp, but we can change this later once we actually have a swamp map.

PR #213, opened 2020-10-03
Sped up swamp and lava bubbles animation.

Dzierzan:


mods/hv/sequences/terrain.yaml
```diff
@@ -20,7 +20,7 @@ icicle2:
 
 lavabubl:
 	active: bits/terrain/lavabubl.png
-		Tick: 100
+		Tick: 60
```
Mailaender, line 23: I think lava is a bit thicker than swamp water.

Dzierzan: This type of animation looks only the best it is fast enough. Sluggish bubble looks just wrong.

PR #219, opened 2020-10-10
Shell casting

Dzierzan:
This adds shell casting for Copter. Initially I also wanted it for scouts, but due to being so small, ejected shells look huge next to them. And if desired, shells can be added for arty units "^DropMagazineCasing". This is also based on Graion changes

shells


mods/hv/rules/aircraft.yaml
```diff
@@ -135,6 +136,9 @@ COPTER:
 		Sequence: propeller
 	WithTurretAttackAnimation:
 		Sequence: shoot
+	GrantConditionOnAttack:
```
GraionDilach, line 139: You don't need this.

mods/hv/rules/vehicles.yaml
```diff
@@ -1,4 +1,4 @@
-# License: CC-BY-SA-4.0
+d# License: CC-BY-SA-4.0
```
GraionDilach, line 1: Typo.

Mailaender:
Can you squash the last 4 commits please?

Mailaender:
I think you pulled in things which made it worse. See https://support.gitkraken.com/working-with-commits/squash/ for help.

Dzierzan:
I did use that page to understand squashing. What doesn't work? I just tested it and it works fine.

Mailaender:
Keep only the first 4 changes and squash the other commits into the 4th.

GraionDilach:
I've did https://github.com/GraionDilach/OpenHV/tree/ShellCasting2 as an attempt. Feel free to merge that one if this is unsalvageable.

PR #220, opened 2020-10-10
Added bridge tiles

### LINKED ISSUE #210: Bridges need proper connection to cliffs
Title^

Dzierzan:
Also edited mount-brosnan map with new bridge tiles. To avoid making a lot of tiles for all cliff and water types, I made sure these new tiles do not have any terrain or water pixels at all.
Closes #210

new bridge


mods/hv/tileset.yaml
```diff
@@ -78,6 +78,10 @@ Terrain:
 		Type: Swamp
 		TargetTypes: Swamp
 		Color: 2B7D38
+	TerrainType@Bridge:
```
Mailaender, line 81: This seems to be unused.

PR #223, opened 2020-10-26
Turn the carryall to a shared unit.

GraionDilach:
This is to fix the balance issue on Tomb Raiders.


Mailaender:
@ABrandau may be working on custom artwork for this.

PR #225, opened 2020-10-30
vscode debugging: build before launching

Phrohdoh:
Like https://github.com/OpenRA/OpenRA/pull/18786 but for OpenHV.
Also fix the configured [cwd] to be the engine dir, so that SDL (and other build artifacts) can be found.
This is equivalent to the `cd "${ENGINE_DIRECTORY}"` in `launch-game.sh`.

[cwd]: https://en.wikipedia.org/wiki/Current_working_directory


Mailaender:
Thanks!

PR #238, opened 2021-01-17
Updated buildings sequences

Dzierzan:
This PR changes how animations are handled. Currently it's a big mess and I do believe animations for production buildings were meant to be used for when something is being built and then finished.

So that means this PR affects Factory2/3, Harbor, Starport1/2 and Base.

There's one thing I couldn't force to work. I wasn't able to force animation "production-done" to play for Factories. I presume those small platforms are somehow colliding with that.


phrohdoh:
Seems there is [a condition that is used (by `base`) but not granted](https://github.com/OpenHV/OpenHV/pull/238/checks?check_run_id=1717908706#step:5:63), `build-incomplete`, which is causing the checks to fail.

Dzierzan:
Yes I know that. But I was too lazy to fix it, besides I need to update buildings-icon art.

PR #250, opened 2021-04-07
Added tanker unit

Dzierzan:
tanker

Currently just replaces tank13. It also has `loaded` sequence but it's not set.

No icon yet, as technician PR hasn't been merged yet.


Mailaender:
Can you rebase this as https://github.com/OpenHV/OpenHV/pull/249 is already merged? This seems to add it a second time.

PR #251, opened 2021-04-10
Added "Doubles" map

Dzierzan:
hv-2021-04-10T201728385Z


Mailaender:
Can you add a license tag?https://github.com/OpenHV/OpenHV/blob/699f9c2378be82413e62f689ffb73770f6ccc224/mods/hv/maps/mount-brosnan/map.yaml#L9

Dzierzan:
Added

PR #255, opened 2021-04-18
Added Civilian Mining Tower

Dzierzan:
image


mods/hv/maps/doubles/map.yaml
```diff
@@ -6,8 +6,6 @@ Title: Doubles
 
 Author: Dzierzan
 
-License: CC-BY-SA-4.0
```
Mailaender, line 9: Please keep this.

mods/hv/rules/buildings.yaml
```diff
@@ -1772,3 +1772,52 @@ TECHBUILDING:
 		Position: Top
 		Margin: 0, 5
 		RequiresCondition: reconstructor > 4
+
+CIVILIANMINER:
```
Mailaender, line 1776: TECHMINER maybe?

PR #261, opened 2021-04-25
AI will build technicians

Dzierzan:
AI will build technicians and `CapturingActorTypes` was lacking technicians as well.

### DIFF
--- a/mods/hv/rules/bots.yaml
+++ b/mods/hv/rules/bots.yaml
@@ -148,6 +148,7 @@ Player:
 			artil3: 5
 			scout1: 1
 			scout2: 1
+			technician: 1
 			missile_tank: 5
 			ship1: 5
 			ship2: 5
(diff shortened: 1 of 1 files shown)


mods/hv/rules/bots.yaml
```diff
@@ -203,8 +205,8 @@ Player:
 	ProductionThrottleBotModule:
 		RequiresCondition: enable-easy-ai
 	PriorityCaptureManagerBotModule:
-		RequiresCondition: enable-hard-ai
-		CapturingActorTypes: tank8
+		RequiresCondition: enable-hard-ai || enable-easy-ai
```
Mailaender, line 208: This was actually on purpose. The easier AI should be more stupid.

mods/hv/rules/bots.yaml
```diff
@@ -185,6 +186,7 @@ Player:
 			artil3: 5
 			scout1: 1
 			scout2: 1
+			technician: 1
```
Mailaender, line 189: Easy AI should not capture anything.

Dzierzan:
I see, maybe in that case I'll close it and you'll do the same on your own.

PR #278, opened 2021-05-03
Trade Platform won't play 'active' animation when it's damaged.

Dzierzan:


mods/hv/rules/buildings.yaml
```diff
@@ -232,6 +232,7 @@ TRADPLAT:
 		StartRepairingNotification: Repairing
 		FinishRepairingNotification: UnitRepaired
 	WithResupplyAnimation:
+		RequiresCondition: !damaged
```
Mailaender, line 235: Do you mean
```suggestion
		RequiresCondition: !burning
```

PR #280, opened 2021-05-04
Added creation sounds

Dzierzan:
- Added 2 new creation sounds, one for base building, the other for actor creation
- Simplified buildings.yaml by moving traits to default.yaml
- Made base turrets and super weapons uncapturable


mods/hv/rules/buildings.yaml
```diff
@@ -1069,13 +1009,10 @@ SILO:
...
-		Condition: build-incomplete
-	WithConstructionBeamOverlay:
-		RequiresCondition: !build-incomplete
-		Condition: beam-incomplete
-	Capturable:
-		RequiresCondition: !build-incomplete && !beam-incomplete
+	-Capturable:
+	-CapturableProgressBar:
+	-CapturableProgressBlink:
+	-CaptureNotification:
```
Mailaender, line 1015: These change look unrelated. Why should the missile silo not be capturable?

Dzierzan: Super weapons are not capturable in CnC games.

Mailaender: It does not make sense to me and I don't like it.

OpenRA.Mods.HV/Traits/Render/WithProductionExitOverlay.cs
```diff
@@ -35,6 +35,9 @@ public class WithProductionExitOverlayInfo : ConditionalTraitInfo, Requires<Rend
 		[Desc("Custom palette is a player palette BaseName.")]
 		public readonly bool IsPlayerPalette = false;
 
+		[Desc("Plays a sound when an actor is spawned.")]
+		public readonly string[] CreationSound = { };
```
Mailaender, line 39: You are adding sound to a render trait which is not the proper way.

Dzierzan:
Alright, I'll close it and seperate that into two PRs.

PR #283, opened 2021-05-04
Added creation sounds

Dzierzan:
Quedicemipez_teleportation.wav is intented to be used for actor creation for Factories and Base. OpenRA doesn't have such trait to play a sound like that.

@Mailaender something for you to code. It will be worth it.


GraionDilach:
VSH has a solution for the top door animation for the Kirov. I don't remember how that works now but using the same condition-granter trait tov enable an AmbientSound could work here.

Dzierzan:
The issue is I would love this sound to be `Player` only, otherwise in let's say 2v2 or 3v3 scenario you will be bombared with teleport sound. Just like you don't hear `BuildUp` for allies or enemies.

PR #284, opened 2021-05-04
Base turret defenses are not longer captureable.

Dzierzan:


Mailaender:
Why? Do you think `PriorityCaptureManagerBotModule` is not smart enough to avoid them?

Dzierzan:
Base defenses shouldn't be capturable, ever!

PR #287, opened 2021-05-08
Added new sounds for Countermeasure Tank

Dzierzan:


mods/hv/bits/audio/cc0/README.md
```diff
@@ -1 +1,3 @@
 Contents of this folder indicate assets licensed under the [CC0 Public Domain Dedication](https://creativecommons.org/publicdomain/zero/1.0/).
+
+* [Sonic1/2/3.wav](https://freesound.org/people/tutarap/sounds/341956/) by kafokafo
```
GraionDilach, line 3: IMO include this in the commit message instead (CC0 does not require even explicit crediting, we only need this info to track them back internally).

Dzierzan: I mean, I know that. But I still wanted to credit the people who created the sound, even if it's CC0.

Mailaender: As a matter of courtesy even CC0 should be credited also as an added proof that it isn't copyrighted.

Mailaender:
This needs a rebase.

Mailaender:
You pulled without rebase. I am not sure what happens when I squash your merge commits.

PR #291, opened 2021-05-09
Added experimental deploy sound for miner actor

Dzierzan:
Please, test that in game.


Mailaender:
It is too short, but the first part kinda fits. Maybe combine it with another one for the part where it is erecting from the ground.

Dzierzan:
Alright, I'll try to improve that.

PR #300, opened 2021-05-13
Volumed down Flamewheel voices

Dzierzan:
Nearly all of them were too loud and not at the same level.


Mailaender:
Compared to the announcer they are quite silent now.

Dzierzan:
I will volume down it too.

PR #301, opened 2021-05-14
Added Sandy Shores map

Dzierzan:
image


mods/hv/maps/sandy-shores/map.yaml
```diff
@@ -0,0 +1,486 @@
+MapFormat: 11
+
+RequiresMod: hv
+
+Title: Dzierzan
+
+Author: Sandy Shores
```
Mailaender, line 7: You flipped author and title.

PR #303, opened 2021-05-14
Added `Chance` parameter for LeaveSmudgeWarhead

Dzierzan:
Also gave LeaveSmudgeWarhead for lightning bolts and added liquid splashes for it.


Mailaender:
The :zap: should probably leave scorches instead. I also don't understand how electricity could add a splash.

Dzierzan:
Lightning is powerful enough to leave craters.

GraionDilach:
RA2 went with scorches only for lightning.

PR #310, opened 2021-05-16
Added Tournament Island map

Dzierzan:
image


Mailaender:
:+1: Don't forget to upload it at https://resource.openra.net/ as well for the server browser preview to work.

PR #312, opened 2021-05-16
Minor map changes

Dzierzan:
Tomb raiders has now tech drill rigs.
Added more stuff for shellmap.


mods/hv/maps/tomb-raiders/map.yaml
```diff
@@ -6,8 +6,6 @@ Title: Tomb Raiders
 
 Author: Graion Dilach
 
-License: CC-BY-SA-4.0
```
Mailaender, line 9: Please don't remove that.

mods/hv/maps/tomb-raiders/map.yaml
```diff
@@ -219,17 +219,17 @@ Actors:
...
 		Owner: Neutral
-		Location: 47,53
-	Actor80: techbuilding
+		Location: 47,52
+	Actor80: techminer
 		Owner: Neutral
-		Location: 47,31
-	Actor81: techbuilding
+		Location: 47,30
+	Actor81: techminer
```
Mailaender, line 231: I think @GraionDilach's whole point of the map was that the tech reconstructors stack up and give more healing.

mods/hv/maps/tomb-raiders/map.yaml
Dzierzan: Goddamn, edit the engine, so it's automatically added or something.

mods/hv/maps/tomb-raiders/map.yaml
Dzierzan: @GraionDilach doesn't mind.

GraionDilach: At that point we had no other tech building. I don't mind this change. When limited eco comes in, we need to revisit the eco on all the amps anyway.

PR #324, opened 2021-08-09
Repurpose shadow pal for crude alpha gradients.

GraionDilach:
image

The second half of the palette is unused incase there are new ideas for the future. These are just the basic 255 0 0/255 128 0 colors afterall.


Mailaender:
There is no use case for it. We may come back to it later.

PR #337, opened 2021-08-24
Adjusted Tank Trap

Dzierzan:
Small adjustment of tank trap prop.

### DIFF
--- a/mods/hv/rules/props.yaml
+++ b/mods/hv/rules/props.yaml
@@ -164,13 +164,21 @@ PROP4:
 	RenderSprites:
 		PlayerPalette: green
 
-TANKTRAP1:
-	Inherits: ^Rock
+TANKTRAP:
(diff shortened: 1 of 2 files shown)


mods/hv/rules/props.yaml
```diff
@@ -164,13 +164,21 @@ PROP4:
...
 	Tooltip:
 		Name: Tank Trap
-
-TANKTRAP2:
-	Inherits: TANKTRAP1
+	-ThrowsShrapnel:
+	-ShakeOnDeath:
+	Explodes:
+		Type: Footprint
+		Weapon: UnitExplodeSmall
```
Mailaender, line 183: I think this is made of steel and not TNT.

Dzierzan: It's a game.

PR #344, opened 2021-08-30
Added new nuclear bomb explosion sound

Dzierzan:


Mailaender:
Yes, that one is a lot less noisy.

PR #349, opened 2021-09-02
Base defense turrets shouldn't realign their guns.

Dzierzan:


Mailaender:
Why?

Dzierzan:
The only RTS game where I've seen base turret realigning is CnC Generals if my memory serves me. Right now they only realign because of one of engine updates changed that and no one budged to update that. 

This is of course my subjective opinion and I am just used to not realigning base turrets.

GraionDilach:
Generals doesn't realign static turrets after a fight. It does have a logic where turrets just rotate randomly during idle.

PR #356, opened 2021-09-06
Added a cruise missile that can be intercepted

Mailaender:
Followup of https://github.com/OpenHV/OpenHV/pull/331.

### DIFF
--- a/OpenRA.Mods.HV/Activities/FlyGuidedIntoTarget.cs
+++ b/OpenRA.Mods.HV/Activities/FlyGuidedIntoTarget.cs
@@ -1,0 +1,202 @@
+#region Copyright & License Information
+/*
+ * Copyright 2021 The OpenHV Developers (see CREDITS)
+ * This file is part of OpenHV, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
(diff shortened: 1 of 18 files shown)


Dzierzan:
Nice, but:

1. Lacks smoke trail for the missile.
2. It weirdly super accelerates when a projectile almost reaches a target.
3. The trajectory could look like more like an arc. Right now it feels like reverse "U" trajectory.
4. A projectile is too fast. Not sure if other missile/rocket projectiles can even reach it.

Mailaender:
1. Added a smoke trail.
2. Decreased target acceleration.
3. That is on purpose as the missile shoots upwards and then locks onto target.
4. I reduced the speed by 15%.

Dzierzan:
Can you move the projectile origin a bit to actor's back so it matches the hole better? Also there's no splash when the missile hits water and probably other fluids as well. I guess this is my oversight, you could update it as well.

Dzierzan:
Hmm... maybe the explosion effect is too small for such missile. Let's use medium one instead.

Mailaender:
Changed it to medium, which is indeed more satisfying.

PR #358, opened 2021-09-06
Balance changes

Dzierzan:
I recommend testing these changes. Exact values you can see on github.

1. Pods are less effective against buildings.
2. Anti-tank turrets have more HP and do a bit more damage against vehicles.
3. Refinery and Miner building have now more HP.
4. Planes are faster and auto attack actors, but ignore buildings.
5. Banshee does more damage against buildings and vehicles to compensate their slow speed.
6. Copters do not less damage against buildings and light vehicles.
7. Artillery turrets do more damage against everything.
8. AA turrets have now the same amount of HP.
9. Increased balloon's and saucer's reveal shroud range by 2 tiles.
10. Sniper pods now one shot kill other pods.

### DIFF
--- a/mods/hv/rules/aircraft.yaml
+++ b/mods/hv/rules/aircraft.yaml
@@ -31,20 +31,21 @@ SHIP2:
 		MuzzleSequence: muzzle
 	AttackAircraft:
 		FacingTolerance: 20
-		PersistentTargeting: false
-		OpportunityFire: false
+		PersistentTargeting: False
 	Aircraft:
 		CruiseAltitude: 2560
 		InitialFacing: 768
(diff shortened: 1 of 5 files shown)


mods/hv/rules/aircraft.yaml
```diff
@@ -31,20 +31,21 @@ SHIP2:
 		MuzzleSequence: muzzle
 	AttackAircraft:
 		FacingTolerance: 20
-		PersistentTargeting: false
-		OpportunityFire: false
+		PersistentTargeting: False
```
Mailaender, line 34: OpenHV style is lowercase

PR #363, opened 2021-09-13
Updated Drop pods power icon

Dzierzan:
Also added "select target" announcer notification when selected.


Mailaender:
Needs a rebase. Sorry.

PR #364, opened 2021-09-13
Added new ship2's artwork

Dzierzan:
Renamed `SHIP2` to `PLANE1`
Sequences and art for previous actor are kept.
plane1
icon_waifu2x_art_noise1_scale_tta_1


mods/hv/rules/bots.yaml
```diff
@@ -88,7 +88,7 @@ Player:
 			scout2: 1
 			technician: 1
 			missile_tank: 5
-			ship1: 5
```
GraionDilach, line 91: Wrong line.

Dzierzan: Thx for heads up.

PR #370, opened 2021-09-14
Added extra sprites for the Force Field building

Dzierzan:


Mailaender:
You forgot about `field.yaml`.

PR #379, opened 2021-09-17
Replaced AA Tank's artwork with new one

Dzierzan:
image


mods/hv/rules/vehicles.yaml
```diff
@@ -88,7 +88,7 @@ TYRIANTANK:
 	RenderSprites:
 		PlayerPalette: green
 
-TANK9:
+AATANK:
```
Mailaender, line 91: Please don't replace it as it works just fine. Do a faction specific alternative instead.

Dzierzan: TANK9 doesn't look like AA tank. Not at all. I prefer to make one more AA tank actor which looks like AA than use sprites which don't indicate that.

PR #380, opened 2021-09-17
Updated damage state sprite for Miner Building

Dzierzan:


Mailaender:
I think we should rather suspend the filling animation when damaged.

Dzierzan:
IMO no, it still needs a clear indication that something is going on.

PR #381, opened 2021-09-17
Small adjustments

Dzierzan:
AA weapons now damage only air units (if an actor was above a building, it was damaged as well, lol).

Adjusted ZOffset for Bombers, their bombs will now appear below them.

Bombers' bombs now explode when they hit trees.

Artillery's projectiles now explode when they hit trees.

### DIFF
--- a/mods/hv/sequences/aircraft.yaml
+++ b/mods/hv/sequences/aircraft.yaml
@@ -149,3 +149,4 @@ landedpod:
 bomber1:
 	idle: bits/bomber1.png
 		Facings: 8
+		ZOffset: 2050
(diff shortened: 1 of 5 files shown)


mods/hv/sequences/aircraft.yaml
```diff
@@ -149,3 +149,4 @@ landedpod:
 bomber1:
 	idle: bits/bomber1.png
 		Facings: 8
+		ZOffset: 2050
```
Mailaender, line 152:
```suggestion
		ZOffset: 2048
```

is the next power of two.

PR #385, opened 2021-09-19
Disabled `OpportunityFire` for planes

Dzierzan:
I actually readded that, as I didn't realize planes would shoot targets while flying above them without ordering the attack.

However if they left alone, they attack foes as intended.


mods/hv/rules/aircraft.yaml
```diff
@@ -32,6 +32,7 @@ PLANE1:
 	AttackAircraft:
 		FacingTolerance: 20
 		PersistentTargeting: false
+		OpportunityFire: False
```
Mailaender, line 35:
```suggestion
		OpportunityFire: false
```

please

PR #387, opened 2021-09-19
Updated bridge tiles with new connectors.

Dzierzan:


mods/hv/maps/antifreeze/map.yaml
```diff
@@ -6,8 +6,6 @@ Title: Antifreeze
 
 Author: SiegeSpud
 
-License: CC-BY-SA-4.0
```
Mailaender, line 9: Don't remove that.

Dzierzan: Goddamn it, it's autoremoved.

Mailaender: I got rid of it completely in https://github.com/OpenHV/OpenHV/commit/c75275c1a7a47580e0866cf6aa6fba0d1ef8beac

PR #388, opened 2021-09-19
Updated bridge tiles with new connectors.

Dzierzan:


Mailaender:
Screenshot_20210919_194030

makes it worse imho.

Mailaender:
Screenshot_20210919_194147

Same here. Brown connects far better:

Screenshot_20210919_194240

Dzierzan:
How come they look worse when brown ones are clearly not supposed to be used like that.

PR #389, opened 2021-09-20
Updated sounds

Dzierzan:
This is an experimental change.

- radio effect has been applied to all unit voices

- announcer and all SFX sounds were normalized, so they should have the same level of volume (might be differences due to nature of some sounds like machine guns)


sources/macros/README.md
```diff
@@ -0,0 +1 @@
+Contents of this folder were created using [Audiacity] (https://www.audacityteam.org/)They are licensed under the [Creative Commons Attribution ShareAlike 4.0](https://creativecommons.org/licenses/by-sa/4.0/) terms.
```
Mailaender, line 1:
```suggestion
Contents of this folder were created using [Audiacity](https://www.audacityteam.org/). They are licensed under the [Creative Commons Attribution ShareAlike 4.0](https://creativecommons.org/licenses/by-sa/4.0/) terms.
```

PR #391, opened 2021-09-25
New announcer voice

Dzierzan:
I was not happy with current "evil-ish" announcer. I recorded new lines plus I added missing ones.


Dzierzan:
I am not super sure about "charging" lines. I think they are not that much needed, in fact I find them a bit annoying.

PR #398, opened 2021-09-30
Disabled sounds for Seed critter

Dzierzan:
It inherited Beast critter sounds.


Mailaender:
Yeah, it doesn't look really talkative.

PR #400, opened 2021-10-02
Added missing gun fire sound for AA boat

Dzierzan:
I dunno if it was always missing or something happened along the way.


GraionDilach:
Why does this one lacks inheriting the default values from `^AntiAirMissile`?

PR #425, opened 2021-11-14
Polished artillery turret

Dzierzan:


mods/hv/rules/buildings.yaml
```diff
@@ -1170,12 +1171,16 @@ ARTILLERYTURRET:
 	AttackOrderPower:
 		PauseOnCondition: disabled
 		Cursor: attack
-		Icon: null
+		#BlockedCursor: howitzer-blocked
```
Mailaender, line 1174: Why is this out-commented? You defined it above.

Dzierzan: Because the engine gives me an error, it is not implemented for that trait. So it is commented out for time being.

PR #427, opened 2021-11-15
Added Synaty Module building

Dzierzan:
image


mods/hv/maps/itrium/map.yaml
```diff
@@ -62,11 +62,6 @@ Actors:
 		Facing: 384
 		Location: 26,34
 		Owner: Synaty Corporation
-	Refinery: refinery
-		Location: 28,40
-		FreeActor: False
```
Mailaender, line 67: This will add a lonely miner.

PR #428, opened 2021-11-16
Updated Silo building

Dzierzan:
image

Also:
Moved Silo to SC
Moved Uplink to YI


Mailaender:
New shadow is actually more detailed:

image

PR #431, opened 2021-11-17
Updated Storages

Dzierzan:
image

This pull request transforms `Storage` building into new Synaty refinery which is later renamed to `Storage` anyway.

Since we no longer use refineries, Yuruki variant is also modified (removing mainly chimneys).

So gameplay wise, players no longer have to bother with building more old silo type buildings (like it was simplified in Red Alert 2).

@Mailaender There are right now unused animations which I want to be used as unload resources animation when a tanker reaches the building. This needs custom trait.


mods/hv/rules/buildings.yaml
```diff
@@ -1520,14 +1463,9 @@ REFINERY:
 		Amount: -30
 	RenderSprites:
 		PlayerPalette: green
-		Image: refinery-yi
+		Image: storage-yi
 		FactionImages:
-			sc: refinery-sc
-	WithResourceStoragePipsDecoration:
```
Mailaender, line 1526: I think we need to keep this as there is no other fullness indicator in the UI.

Dzierzan: Wait, I thought we are dropping silos mechanic? Like it is in Red Alert 2, no more silos. So why pips are needed for that?

Mailaender: Well, but you kept

```salt
	StoresResources:
		Capacity: 5000
```

Dzierzan: Ohh snap!

PR #434, opened 2021-11-17
Updated Turret2

Dzierzan:
This pull request is meant to fix the issue with double shooting sound for turret2 which makes it louder than indented. So I created a new muzzle flash and made the turret using one barrel when it shoots (just like Banshee does it now).

Since I created the new muzzle flash, I wanted to adjust its offsets for each facing, as I did it for other actors, however for this one, the game crashes, saying that the spritesheet uses 15 frames, but there's only one.

I think I checked everything I could, PNG metadata, yaml code. Just nothing. Heck, I tested the yaml code for another actor (a mobile one) and it worked just fine, but the same yaml code used for another base turret actor crashes the game. Something's fishy here.

(commit names are subject to be changed, it's a draft for now).


Dzierzan:
OK, I decided to make a new art for that turret, one barrel is enough and even having one barrel caused crash if I wanted to have seperate muzzle flash and offsets for each facing. So I decide to hardcode muzzle flash with turret spritesheet.

image

PR #441, opened 2021-11-20
Added extra announcer voices

Dzierzan:
I recorded extra announcer voices for super powers, when they are incoming, used or detected. However I couldn't find notifications tokens for player when enemy super power is really detected/built. I know Red Alert 2 game did have that, so I thought OpenRA would have it too. Did I miss something?

I also re-recorded `airstrike` voices, because I realized I pronounced `strike` part wrong.


Mailaender:
> I know Red Alert 2 game did have that, so I thought OpenRA would have it too. Did I miss something?

I just added them in https://github.com/OpenRA/OpenRA/pull/19811

PR #443, opened 2021-11-20
Brought back Base2

Dzierzan:
Matze, when you changed all filenames and code to use generic names for building actors, you accidently removed `FactionImages` for the Base, so SC variant would be never rendered.


Mailaender:
It is more than a faction image due to the lighting.

PR #455, opened 2021-11-25
Updated palette

### LINKED ISSUE #454: Wrong purple color range?
I've noticed long time ago that there's something wrong with the purple color in game, it feels washed out. So I took a screenshot and compared it with the raw sprite. The result is below. @Mailaender are you sure you are using correct palette? I believe Daniel Cook created several palettes, for example snow trees have a bit different color range and I do believe I've seen palette with washed out purple colors as well.

purple

Dzierzan:
Purple color range was wrong, which made those colors washed out. I updated it with proper color range.

Closes #454


mods/hv/bits/colors.pal
```diff
@@ -1,259 +1,259 @@
...
+0 255 0
+0 255 0
+0 255 0
+0 255 0
+0 255 0
+0 255 0
+0 255 0
+0 255 0
+0 255 0
+0 255 0
```
Mailaender, line 163: Removing the colors reserved for details is pretty broken:

image

Dzierzan: Right, I forgot mothership uses full palette. Let me fix that

mods/hv/bits/colors.pal
```diff
@@ -1,259 +1,259 @@
-JASC-PAL
-0100
-256
-0 0 0 140
```
Mailaender, line 4: This has an impact on some shadows:
image

Dzierzan: I am not sure why... I reapplied the palette for this shadow spritesheet, but there was no effect. Is there something else going on here for this particular shadow?

GraionDilach: For starters, `WithIdleOverlay@Shadow` is only defined on `^DefenseBuilding` for some reason in `defaults.yaml`. This might mean that a fair amount of the shadows might use wrong palette. Also keep in mind that "TS-styled" split shadows (which the Mining Tower actually uses) are forced to share the palette between all frames. This is also one of the limitations of the `WithSpriteAnimation` system.

mods/hv/bits/colors.pal
```diff
@@ -1,259 +1,259 @@
...
-68 51 59
-76 58 66
-83 65 73
-91 73 81
-98 81 88
-105 89 96
-113 98 104
-120 107 112
-128 116 121
+JASC-PAL
```
GraionDilach, line 1: What is going on here? Have you accidentally changed the EOL character?

Dzierzan: I literally took Mothership's palette and update purple range and saved it in Aseprite.

mods/hv/bits/colors.pal
Mailaender: Just don't remove the `140` entry.

Dzierzan:
It should be OK now

PR #457, opened 2021-11-26
Updated Mother building artwork

Dzierzan:
mother


mods/hv/maps/shellmap/map.yaml
```diff
@@ -686,11 +679,6 @@ Actors:
 		Location: 88,22
 		Faction: Random
 		Owner: Synapol Corporation
-	Actor246: storage
-		Location: 69,13
-		Faction: Random
-		Owner: Synapol Corporation
-		FreeActor: False
```
Mailaender, line 693: This looks like an accidental change.

Dzierzan: Ohh, not, not, I actually not, building was wrong. I mean YI storage was in a place of SC storage. But I again forgot to disable FreeActor for that. Let me fix it.

Mailaender: This is still `storage` though, you want `storage2`?

Dzierzan: No such actor as `storage2` ^^. It uses:

`	RenderSprites:
		PlayerPalette: green
		Image: storage2
		FactionImages:
			sc: storage`

Mailaender: Yes, but this is identical except for `Faction: Random` is that the actual wanted change here?

Dzierzan: And what's the issue? Considering nearly every actor has `Faction: Random` in Shellmap?

PR #458, opened 2021-11-26
Removed tank4

Dzierzan:
IMO this actor was never finished/polished by Daniel Cook and we already replaced with my Radar Tank.

Plus tank4 was still used for `DeployableActorTypes`, where it shouldn't.


mods/hv/sequences/vehicles.yaml
```diff
@@ -72,18 +72,6 @@ radartank:
 	icon: bits/units-icons.png
 		Start: 47
 
-tank4:
-	idle: bits/tank4.png
```
Mailaender, line 76: I don't want it removed. There is still a role for it in missions.

PR #461, opened 2021-11-28
Added small walls

Dzierzan:
image

No reverse make animation for walls when you sell them because wall trait doesn't like `WithWallSpriteBody`.


mods/hv/sequences/buildings.yaml
```diff
@@ -1276,6 +1276,22 @@ wall:
 		Start: 1
 	icon: bits/icon.png
 
+wallsmall:
```
Mailaender, line 1279:
```suggestion
barrier:
```

maybe?

mods/hv/rules/buildings.yaml
```diff
@@ -1775,6 +1775,87 @@ HARBOR2:
...
+		BuildPaletteOrder: 80
+		Queue: Defense
+		Prerequisites: generator, module
+		Description: Basic wall defense.
+	RequiresBuildableArea:
+		AreaTypes: building
+		Adjacent: 4
+	MustBeDestroyed:
+		RequiredForShortGame: false
+	RepairableBuilding:
```
Mailaender, line 1796: Really?

Dzierzan: Normally I'd use `Inherits: ^BaseBuilding` to have less code here, but I had one tiny issue. I couldn't get rid of `		Notification: BuildingSold` because even if I added it as empty, it was still inherited, so the sound notification was used.

Dzierzan: I considered adding null sound, but then you'd complain why on Earth I added it! ^^

Mailaender: I mean how should walls regenerate and should they from a gameplay perspective?

PR #463, opened 2021-11-30
Added gates

Dzierzan:
It's not finished PR yet. A solution needs to be found for better connection with wall segments.


mods/hv/rules/buildings.yaml
```diff
@@ -1852,6 +1852,106 @@ BARRIER:
...
+	WithConstructionBeamOverlay:
+	UpdatesPlayerStatistics:
+		AddToAssetsValue: false
+	WithShadow:
+		Offset: 200,200,0
+		ZOffset: -1000
+	Sellable:
+		SellSounds: Quedicemipez_teleportation-barrier-sell.wav
+
+GATE-A:
```
Mailaender, line 1921: I would prefer something a bit more descriptive like
```suggestion
GATE-EW:
```

Mailaender: or maybe `VGATE/HGATE`

mods/hv/rules/buildings.yaml
```diff
@@ -1852,6 +1852,106 @@ BARRIER:
...
+	Building:
+		RequiresBaseProvider: true
+		TerrainTypes: Clear, Road, Crater, Grass, Grass Pit, Snow, Red Snow, Sand, Stone, Tech
+		AllowInvalidPlacement: false
+		BuildSounds: Quedicemipez_teleportation-barrier.wav
+	Buildable:
+		BuildPaletteOrder: 15
+		Queue: Defense
+		Prerequisites: module
+		Description: Automated barrier that opens for allied units.
```
Mailaender, line 1867:
```suggestion
		Description: Automated barrier that opens for allied units.
	Replacement:
		ReplaceableTypes: wall
```

mods/hv/rules/buildings.yaml
```diff
@@ -1852,6 +1852,106 @@ BARRIER:
 	-Selectable:
 	Interactable:
```
Mailaender, line 1853:
```suggestion
	Interactable:
	Replaceable:
		Types: wall
```

PR #465, opened 2021-12-02
Added shadows for walls and gates.

Dzierzan:
image

Stupid overlapping shadow issue... I had to fight against it. Anyway, wall shadows connect nicely in 99% cases. If you place pillars next to each other, you might notice small overlapping shadow spot. Example below:

image

Sadly, gates' shadows don't work so nicely with walls. Technically speaking it can be just ignored. The pro is that now shadow moves together with the gate when it opens and `open` sprite no longer casts shadow, which looked like it was floating when `WithShadow` trait was used.

gateshadows


Dzierzan:
Improved, but gates alone look weird now. Spritesheets contain now shadows for gates if they were built alone. No idea how to implement them, though.

image

PR #466, opened 2021-12-02
Adjusted RevealsShroud for a lot of actors

Dzierzan:
With these changes, whoever sees enemy first, has an advantage. Basically weapon range for units is longer than they see. So if you use scouts like Balloons or Radar Tank, you can suprise your enemy,

Also adjusted vision for some buildings, expecially tech ones, those were huge for some reason.

Also removed `MinRange`, I've got no idea what's the point of that. I know that it was brought from OpenRA RA, but I don't get the idea of units having minimal vision range.

### DIFF
--- a/mods/hv/rules/buildings.yaml
+++ b/mods/hv/rules/buildings.yaml
@@ -201,10 +201,10 @@ GENERATOR:
 		Type: Steel
 	RevealsShroud:
 		Range: 4c0
-		MinRange: 4c0
+		MinRange: 2c0
 		RevealGeneratedShroud: False
 	RevealsShroud@Hacked:
-		Range: 4c0
+		Range: 2c0
 	Power:
 		Amount: 130
 	Targetable:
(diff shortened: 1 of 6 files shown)


Mailaender:
> Also removed `MinRange`, I've got no idea what's the point of that.

It is required for the hacker tank to work.

Dzierzan:
`MinRange` is back

mods/hv/rules/buildings.yaml
```diff
@@ -255,13 +253,11 @@ RADAR:
 		Type: Steel
 	RevealsShroud:
 		Range: 10c0
-		MinRange: 4c0
```
Mailaender, line 258: `MinRange` is not back for the buildings.

Dzierzan: I forgot, right

mods/hv/rules/vehicles.yaml
```diff
@@ -902,9 +867,12 @@ TANK8:
 	RenderSprites:
 		PlayerPalette: green
 	RevealsShroud:
-		Range: 6c0
+		Range: 7c0
+		MinRange: 3c0
 	CreatesShroud:
-		Range: 8c0
+		Range: 9c0
+	RevealsShroud@Hacked:
```
Mailaender, line 874: Hacker tank can get hacked. Alright. I take it.

PR #474, opened 2021-12-05
Added `DetectCloaked` for actors

Dzierzan:
- Base defenses detect clocked actors at 5 tiles range

- Vehicles at range 1 (Hacker and countermeasure tank at 2)

- Radar tank has the same detecting range as far as it reveals shroud.

- Pods at range 1 (Sniper at 2)

### DIFF
--- a/mods/hv/rules/vehicles.yaml
+++ b/mods/hv/rules/vehicles.yaml
@@ -257,15 +257,18 @@ RADARTANK:
 		RequiresCondition: !undeployed && real-actor
 		Sequence: extended
 		Name: deployed
-	DetectCloaked:
-		RequiresCondition: !ecmdisabled && deployed
-		Range: 12c0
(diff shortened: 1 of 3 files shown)


mods/hv/rules/vehicles.yaml
```diff
@@ -257,15 +257,18 @@ RADARTANK:
...
 	RenderDetectionCircle:
 		TrailCount: 3
 	WithFacingSpriteBody:
 		RequiresCondition: undeployed
 	RenderSprites:
 		PlayerPalette: green
+	DetectCloaked:
+		Range: 8c0
+		RequiresCondition: !ecmdisabled && undeployed
+	DetectCloaked@Undeployed:
```
Mailaender, line 269:
```suggestion
	DetectCloaked@Deployed:
```

PR #480, opened 2021-12-28
Updated terrain actors

Dzierzan:
This change will make overlapping for terrain actors like trees less visable:

Before:
image
After:
image


Mailaender:
This one looks unchanged:

image

Dzierzan:
It's a tile, not actor

PR #486, opened 2021-12-29
Removed artil2 unit

Dzierzan:
It was unused and I didn't like the work I did for this unit.


mods/hv/sequences/vehicles.yaml
```diff
@@ -213,17 +213,6 @@ artil:
...
-artil2:
-	idle: bits/artil2.png
-		Facings: 8
-	muzzle: bits/artil2.png
-		Start: 8
-		Length: 3
-		Facings: 8
-		Tick: 100
-	icon: bits/units-icons.png
-		Start: 41
```
Mailaender, line 225: You could remove the icon as well.

Mailaender:
Symmetry!

PR #487, opened 2021-12-29
Updated shellmap

Dzierzan:
Two generators were wrong


Mailaender (commented):
I don't understand this change. It only seems to change the actor ID and shift the values around.

Dzierzan:
I don't get it.

I removed these generators and placed correct ones, but I just looked at that shellmap (yes, I made sure I am using the right branch), looks totally unchanged.

image

Dzierzan:
`Faction: Radom` broke that. Without it works as intended.

PR #502, opened 2022-01-04
Added sounds for seed critter

Dzierzan:


mods/hv/rules/animals.yaml
```diff
@@ -90,8 +90,9 @@ SEED:
 		WanderMoveRadius: 1
 		MinMoveDelay: 0
 		MaxMoveDelay: 0
-	-Voiced:
-	-DeathSounds:
+	Voiced:
+		VoiceSet: SeedVoice
+	DeathSounds:
```
Mailaender, line 95: It should be already added. See the removal above.

Dzierzan: Sorry, I don't get it. What should be already added? `DeathSounds`?

Mailaender: Remove `DeathSounds:` and keep `-DeathSounds:` removed.

PR #519, opened 2022-01-22
Added patrol boat

Dzierzan:


mods/hv/rules/ships.yaml
```diff
@@ -47,6 +47,58 @@ BOAT3:
...
+		Cost: 400
+	Tooltip:
+		Name: Patrol boat
+		GenericName: Boat
+	Buildable:
+		Queue: Ship
+		BuildAtProductionType: Ship
+		Prerequisites: ~harbor.yi
+		BuildPaletteOrder: 40
+		Description: Light patrol boat.\nExcellent against air units.\nNot effective against other units.\nCan detect submarines.\n
```
Mailaender, line 63: Aren't submarines the natural counter to submarines?

mods/hv/rules/ships.yaml
```diff
@@ -47,6 +47,58 @@ BOAT3:
 	WithMuzzleOverlay:
 		IgnoreOffset: True
 
+AABOAT:
```
Mailaender, line 50: I would name it differently as I think its role could change in the near future.

mods/hv/rules/ships.yaml
Dzierzan: Well, I can make submarines detect other submarines and this unit won't do so. In OpenRA RA nearly everything can detect submarine at close range, apart from Battleships.

mods/hv/rules/ships.yaml
Dzierzan: Sure.

mods/hv/rules/ships.yaml
Mailaender: I mean it can detect them, but it does only little damage which I think is confusing.

Dzierzan: Well, you are supposed to use cheap patrol boats and submarines together.

PR #520, opened 2022-01-23
Updated decorations

Dzierzan:
- Added cactus
- Added snow tree
- Added grass tree
- Snow and grass trees should leave mud/ash on any terrain.

mud-ash2


mods/hv/rules/terrain.yaml
```diff
@@ -207,31 +238,48 @@ TreeAsh15:
...
+	Immobile:
+		OccupiesSpace: false
+	-MapEditorData:
 
-TreeMudSnow16:
-	Inherits: ^TreeMud
-	LaysTerrain:
-		Offset: 0,1
-		Template: 1432
+TREEASH2:
```
Mailaender, line 250:
```suggestion
TREEASH:
```

There is no `TREEASH` actor.

PR #525, opened 2022-01-25
Updated submarine

Dzierzan:
I made the submarine a bit longer (1px)


Mailaender:
Why?

Dzierzan:
Reasons beyond your comprehension! Nah, I just felt like such powerful unit needs to be a bit longer.

PR #544, opened 2022-05-01
Added new voice for hackerpod

Dzierzan:
Also enabled hackerpod production for AI. It has to be taught how to use it, though.


Mailaender:
First it needs to learn how to use it, then it can be safely enabled.

PR #545, opened 2022-05-01
Updated ZOffset for rocks

Dzierzan:


Mailaender:
Choosing the same Z offset like used for the shadows seems wrong.

mods/hv/sequences/terrain.yaml
```diff
@@ -16,6 +16,7 @@ rock1:
 
 rock2:
 	idle: bits/terrain/rock2.png
+		ZOffset: -2000
```
Mailaender, line 19: This is not a power of 2 which might cause problems in zooming. Maybe `1024`.

Issue #551, opened 2022-05-03
Some maps have imbalanced resources


WilliamWeberBerrutti:
(for some reason I can't add screenshots to illustrate better)

(also I'm not sure if here's the best place to do this)

Fire Souls:

It has extra resources on the top of the left half, but not on the bottom of the right half.

Tomb Raiders:

One resource deposit cannot be harvested on the central left "square" with a pyramid in the middle. Also all resources positioning are not as close as possible to the starting building. Maybe the design is residual from an old version of the game?

GraionDilach:
Regarding Tomb Raiders, it was badly updated when the tower anchor was moved from left to right, so yes.

PR #554, opened 2022-05-03
Fixup central mining positions on Tomb Raiders.

GraionDilach:
Related to #551.


GraionDilach:
Decided to sacrifice balance for visuals. Due to the map being mirrored on both x and y axis, there's no good option here. I'll stick to the original deepsink resource alignment.

PR #567, opened 2022-06-05
Added fire truck for FireWatch map

Dzierzan:


mods/hv/maps/firewatch/map.yaml
```diff
@@ -41,6 +41,16 @@ Actors:
 	Actor0: base2
 		Owner: FireBrigade
 		Location: 35,34
+	Actor2: watertank
+		Location: 33,39
+		Faction: yi
+		Facing: 384
+		Owner: FireBrigade
+	Actor3: firetruck
```
Mailaender, line 49: Can you please remove these test actors?

PR #589, opened 2022-08-02
Added transport aircraft for drop zone.

Dzierzan:
transportaircraft


mods/hv/rules/aircraft.yaml
```diff
@@ -562,3 +562,40 @@ CARGOSHIP:
 		Type: CenterPosition
 	DamageMultiplier@Invulnerable:
 		Modifier: 0
+
+TRANSPORTAIRCRAFT:
```
Mailaender, line 566:
```suggestion
AIRLIFTER:
```
?

PR #593, opened 2022-09-16
Rebalance Pods and Vehicles

WilliamWeberBerrutti:
The motivation for these changes is to create an unique gameplay for OpenHV and improve balancing. I think I didn't forget anything.

I couldn't test too much with other players so it needs more testing, specially on the high level play.

The list of changes are as follows:

- LOS increased for pods and vehicles
- **Resource Transporter:** increased capacity from 8 to 9 so it gives a bit more resources when delivering to the Storage Facilities
- **Synapol Helicopter's Chaingun:** reduced damage of None from 180 to 160 and Light from 40 to 35
- **Flame Pod's Flamethrower:** reduced damage of None from 90 to 45 and Steel from 45 to 22
- **Elite Pod's Heavy Machinegun:** reduced damage of None from 200 to 100 and Steel from 35 to 20. Now it's an air support pod. The idea is to be used as auxiliary damage unit with other pods and vehicles

### DIFF
--- a/mods/hv/rules/defaults.yaml
+++ b/mods/hv/rules/defaults.yaml
@@ -183,11 +183,11 @@
 		CrushClasses: Pods
 		WarnProbability: 15
 	RevealsShroud:
-		Range: 4c0
-		MinRange: 2c0
+		Range: 8c0
+		MinRange: 4c0
 		RevealGeneratedShroud: False
 	RevealsShroud@Hacked:
-		Range: 2c0
+		Range: 8c0
 	DetectCloaked:
 		Range: 1c0
 
--- a/mods/hv/rules/pods.yaml
+++ b/mods/hv/rules/pods.yaml
@@ -103,7 +103,7 @@ SNIPERPOD:
 		Speed: 70
 		Locomotor: pod
 	RevealsShroud:
-		Range: 8c0
+		Range: 10c0
 	Armament@Primary:
 		Weapon: SniperRifle
 		MuzzleSequence: muzzle
@@ -246,7 +246,7 @@ ELITEPOD:
 	Inherits@AutoTarget: ^AutoTargetGroundAssaultMove
 	Buildable:
 		Queue: Scouts
-		Description: Elite airborn pod.\nArmed with heavy machine gun.\n  Strong vs Pod, Light Vehicles and Buildings.
+		Description: Elite airborn pod.\nArmed with heavy machine gun.\n  Strong vs Pod and Light Vehicles\n  Weak vs Buildings.
 		Prerequisites: ~module, techcenter
 	Valued:
 		Cost: 750
--- a/mods/hv/rules/vehicles.yaml
+++ b/mods/hv/rules/vehicles.yaml
@@ -22,11 +22,11 @@ TANK3:
 	Armor:
 		Type: Heavy
 	RevealsShroud:
-		Range: 6c0
-		MinRange: 3c0
+		Range: 8c0
+		MinRange: 4c0
 		RevealGeneratedShroud: False
 	RevealsShroud@Hacked:
-		Range: 3c0
+		Range: 4c0
 	Turreted:
 		TurnSpeed: 25
 		Offset: -125,0,50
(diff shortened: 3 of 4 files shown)


mods/hv/rules/defaults.yaml
```diff
@@ -183,11 +183,11 @@
...
 		WarnProbability: 15
 	RevealsShroud:
-		Range: 4c0
-		MinRange: 2c0
+		Range: 8c0
+		MinRange: 4c0
 		RevealGeneratedShroud: False
 	RevealsShroud@Hacked:
-		Range: 2c0
+		Range: 8c0
```
Mailaender, line 190:
```suggestion
		Range: 4c0
```

Dzierzan:
- **_LOS increased for pods and vehicle_** - The reason why I made vision kinda low is because this makes scouts like Radar Tank or Balloon more usable. Notice that right now units can shoot further than they see. That means whoever sees the target first, has advantage.

- _**Synapol Helicopter's Chaingun**_ - why not. It is kinda OP right now. Small nerf won't hurt.

- _**Flame Pod's Flamethrower**_ - IMO no. It's a short range unit. And it's quite fragile (when it explodes, it damages other units). It needs to be powerful, so it can dispose buildings or pods as fast as possible, while still being fragile. On the other hand, I think I made a mistake blocking Flame Pod and Mortar pod behind Factory... I still dunno...

- _**Elite Pod's Heavy Machinegun**_ - I've got missed feeling about this. At first I did want to have something like Red Alert 2 Rocketeer, but then I thought making them "ultimate" pod would be more interesting, where it would replace machine gun pods. Up to the debate.

- **_Resource Transporter_** - hmmm... hard to say. I say neither yes nor no.

WilliamWeberBerrutti:
Sorry for the delay! Here are my thoughts:

- **LOS increased for pods and vehicles:** I see, but will the scouts be not usable even with these changes (since they have a huge LOS)? For this change, I took inspiration from Starcraft: there the units do not have as restrict LOS as in RA, but some have less LOS than their actual range (think about Terran Siege Tanks). Also I added their LOS to make it more unique compared to RA, but maybe the LOS range is too high. The proposed LOS increase can be reduced a bit too. Also, I personally don't miss scout units when I go air, but sometimes I make them to keep an eye on strategic locations on the map so I can see where the enemy army is moving.
- **Resource Transporter:** Yea this change is hard to define if it's good or not without some gameplay. I just did this since the resources seemed too scarce on a previous version, even if the player had a macro-oriented approach. It's probably unnecessary.
- **Synapol Helicopter's Chaingun:** nice!
- **Flame Pod's Flamethrower:** What if this unit didn't do splash damage when destroyed? Also, since this unit and Mortar pod are powerful, maybe it's a good idea for them to require Factory or maybe the Radar Dome.
- **Elite Pod's Heavy Machinegun:** The question is: imagine late game, what should be the unit composition if this unit isn't nerfed? If the player has Barracks, it'll keep spamming Elite Pods probably, including other air units. How to counter this mass of air units? With the Missile Vehicle only, since the Missile Pod is too weak vs them, but I suppose it's hard to produce enough Missile Vehicles to counter this mass, so it has potential for an unstoppable army composition which can snowball easily.

Dzierzan:
Elite pods were nerfed already. Maybe a bit too much, though. I will be checking that. 

About pods and vehicles. I am actually reconsidering that. I will be testing new LOS soon.

If I remember correctly, copter was nerfed. 

Tankers now deliver more resources.

Flamers/mortars. Well, there has to be reason to either build pods or vehicles. Flamers or mortars should be base or pods killers, otherwise you will always choose vehicles.

WilliamWeberBerrutti:
I just played some matches with some people and indeed resources are more scarce. We just begun mining and then after a few minutes the announcer said "Resources depleted", so there's something wrong with the resources at the moment it seems.

And about the Helicopter: at the moment it seems too powerful and it needs to be nerfed. There was a player which spammed Missile Pods and Missile Vehicles, and he couldn't beat my army composed by Helicopters. I also tried with Helicopters and Elite Pods. Maybe both need nerfing.

PR #670, opened 2022-11-23
Re-add Dropship husk with fix

SeanScherer:
Readds the Dropship husk (ie crash animation on death), but with fix so that its not triggered when landed ("RequiresCondition: airborne").

I thought its a shame to lose the crashing animation, so I had a look if #660 can't be fixed (while keeping the husk/animation).

Properly fixes https://github.com/OpenHV/OpenHV/issues/660.


SeanScherer:
Tested against the [latest pre-release](https://github.com/OpenHV/OpenHV/releases/tag/20221119) .

Mailaender:
Even works well when there is a tank hooked up.

PR #708, opened 2022-12-03
Increased ammo pool for aircraft units.

Dzierzan:


mods/hv/rules/aircraft.yaml
```diff
@@ -267,15 +267,15 @@ BANSHEE:
 		LocalOffset: 700,200,0, 700,-200,0
 		PauseOnCondition: !ammo
 	AmmoPool:
-		Ammo: 6
+		Ammo: 12
```
Mailaender, line 270: One volley can bring down almost any building.

Dzierzan: It's not supposed to? After destroying say generator its ammo is depleted. It has to have some uses.

Dzierzan: Ammo pool still makes it nerfed as you can't continuously attacking. But the first blow needs to be effective.

PR #712, opened 2022-12-03
Updated shellmap.

Dzierzan:


mods/hv/maps/shellmap/map.yaml
```diff
@@ -26,12 +26,12 @@ Players:
 		Name: Creeps
 		NonCombatant: True
 		Faction: Random
-	PlayerReference@Yuruki:
+	PlayerReference@Yuruki Industries:
```
Mailaender, line 29: I guess this is an automated change.

Mailaender:
image

mods/hv/maps/shellmap/map.yaml
Dzierzan: Yes

Dzierzan:
I forgot about these two.

PR #735, opened 2023-04-05
Removed `generator` prerequisite from first buildings.

Dzierzan:
- Player can now build straight away `Generator`, `Module` or `Storage`
- No more mandatory `Generator`. You will have to build it anyway later if you want to provide extra power.
- Main base provides 120 power now.

### DIFF
--- a/mods/hv/rules/buildings.yaml
+++ b/mods/hv/rules/buildings.yaml
@@ -36,7 +36,7 @@ BASE:
 		CircleBorderWidth: 4
 	WithBuildingPlacedAnimation:
 	Power:
-		Amount: 0
+		Amount: 120
 	ProvidesPrerequisite@BuildingName:
 	ProvidesPrerequisite@Yuruki:
 		Prerequisite: structures.yi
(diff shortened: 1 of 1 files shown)


Mailaender:
This requires changes to the tutorial misssions.

mods/hv/rules/buildings.yaml
```diff
@@ -285,7 +285,7 @@ RADAR:
 	Buildable:
 		Queue: Building
 		BuildPaletteOrder: 90
-		Prerequisites: generator, storage, ~structures.yi
+		Prerequisites: base, storage, ~structures.yi
```
Mailaender, line 288: `Prerequisites: base` is redundant as you need a build queue.

mods/hv/rules/buildings.yaml
```diff
@@ -1537,7 +1537,7 @@ STORAGE:
 	Buildable:
 		BuildPaletteOrder: 35
 		Queue: Building
-		Prerequisites: generator
```
Mailaender, line 1540: This dependency should stay to not completely disrupt the tech tree.

Mailaender:
Have you tested how the AI reacts to this?

Dzierzan:
Yes, I tested it with AI. There are not problems with it.

Dzierzan:
OK, after leaving `Generator` prerequisite for Storage AI was always starting with Module and running out of money since it was training pods and building defenses. If I again take away that prerequisite, everything comes back to normal.

PR #737, opened 2023-04-10
Update buildings.yaml

SiegeSpud:
Main base to accept tankers as a backup for storages.


Dzierzan:
IMO no. If you destroy enemy's storage, then you cut off their resource supply line. With that change, there's basically no stress of losing storage.

PR #758, opened 2023-08-05
Allowed Naval Transport to travel ice cliff tiles.

Dzierzan:
Moreover:
- Naval Transport can hold now two vehicles.
- Land Transport and Dual Artillery cannot Naval Transport enter because they are visually too huge.
- Water trail will be only spawned on water tiles (it was also spawned on beach tiles which looked odd).

### DIFF
--- a/mods/hv/rules/ships.yaml
+++ b/mods/hv/rules/ships.yaml
@@ -386,9 +386,9 @@ NAVALTRANSPORT:
 	DetectCloaked:
 		Range: 1c0
 	WithLandingCraftAnimation:
-		OpenTerrainTypes: Shore
+		OpenTerrainTypes: Shore, IceCliff
 	Cargo:
-		Types: Scout
+		Types: Scout, Vehicle
 		MaxWeight: 10
 		AfterUnloadDelay: 40
 	WithCargoPipsDecoration:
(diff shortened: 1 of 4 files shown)


mods/hv/rules/ships.yaml
```diff
@@ -386,9 +386,9 @@ NAVALTRANSPORT:
 	DetectCloaked:
 		Range: 1c0
 	WithLandingCraftAnimation:
-		OpenTerrainTypes: Shore
+		OpenTerrainTypes: Shore, IceCliff
```
Mailaender, line 389: Not on cliffs. Like I said, maybe on a new terrain type `IceFloes`.

Dzierzan: Not sure how you want to achieve that without making new tiles.

Mailaender: Change the terrain type for tiles 269-294 to something new: `IceFloes`.

Dzierzan: But I wanted to use for already existing 3 ice maps with large water areas and none of them use those proposed tiles in huge numbers. I really think for gameplay sake you lose nothing here.

Mailaender: You can also edit the maps and add some ice floe shores, as they mostly predate the water transport. I think it is a good idea to have traversable and non-traversable terrain here to give the mappers a choice like we do for temperate terrain tiles.

Mailaender:
Can you please put everything in separate commits?

Dzierzan:
I don't see such option in GitKraken. Is it needed? Or I'd have to start anew?

PR #760, opened 2023-08-05
Added deploy animation for Water Minelayer.

Dzierzan:


Mailaender:
The mine reappears so fast that you can barely see it was missing.

Dzierzan:
You want more delay? The catch is you can interrupt the animation, so even if I add more hardcoded delay to the animation itself, it will look bad then you interrupt, at least at the beginning.

Dzierzan:
I will redesign the water minerlayer.

PR #767, opened 2023-08-20
Add LoadCargoBotModule for AI cargo control

### LINKED ISSUE #721: Bots don't know how to transport units
Neither cargo nor carryall.

dnqbob:
Add LoadCargoBotModule as I said in #721. Closes #721

Compared to the one in SP I add a simple unload logic, for better unload control I still recommend to use the [AutoDeployer.cs](https://github.com/MustaphaTR/OpenRA/blob/rv-engine/OpenRA.Mods.AS/Traits/AutoDeployer.cs) and [AutoDeployManager.cs](https://github.com/MustaphaTR/OpenRA/blob/rv-engine/OpenRA.Mods.AS/Traits/Player/AutoDeployManager.cs) from AS, unless you guys have better ideas.

### DIFF
--- a/mods/hv/rules/bots.yaml
+++ b/mods/hv/rules/bots.yaml
@@ -98,6 +98,7 @@ Player:
 			aatank: 3
 			tank15: 1
 			tank16: 1
+			transprt: 1
 			artil: 5
 			scout1: 1
 			scout2: 1
@@ -154,6 +155,10 @@ Player:
 	CubePickupBotModule:
 		RequiresCondition: enable-rogue-ai
 		ExcludedUnitTypes: miner, tanker1, tanker2, hackerpod
+	LoadCargoBotModule:
+		RequiresCondition: enable-rogue-ai
+		TransportTypes: transprt
+		PassengerTypes: mortarpod, flamepod, scout1, scout2
 	DeployActorBotModule:
 		RequiresCondition: enable-rogue-ai
 		DeployableActorTypes: radartank, hackerpod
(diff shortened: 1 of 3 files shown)


mods/hv/rules/bots.yaml
```diff
@@ -98,12 +98,14 @@ Player:
...
 			tank15: 1
 			tank16: 1
+			transprt: 1
 			artil: 5
 			scout1: 1
 			scout2: 1
 			mortarpod: 1
 			flamepod: 1
 			elitepod: 1
+			sniperpod: 1
```
Mailaender, line 108: This is a fix worthy of its own commit.

dnqbob: Separated.

Mailaender:
Units can shoot from vehicles so we don't need a sophisticated eject on attack. It would actually be counter produtive. This works pretty well. The only problem I saw is that it builds too many and sends empty transport to the front which are defenseless here, but I think the API does not have a solution for that. I will probably just limit the amount of transports being built.

dnqbob:
> The only problem I saw is that it builds too many and sends empty transport

the PR https://github.com/OpenRA/OpenRA/pull/20892#discussion_r1216024028 can solve it. It is due to UnitBuilder build unit compeletely randomly before reach the idle units limit.

dnqbob:
> Units can shoot from vehicles so we don't need a sophisticated eject on attack.

But if we want apply this to water transport or air transport, we will need that.

PR #788, opened 2023-09-07
More consistent, remapable land mines added

SiegeSpud:
More consistent, remapable land mines added based on graphic by Daniel Cook. New land mines are more similar to water mines. Added SiegeSpud as additional author.


Mailaender:
You need to run `./utility.sh --png-sheet-import ../mods/hv/bits/mines.png` to bake the metadata into the sprite. There is also a batch script for Windows that works similar.

Dzierzan:
I'd squish commits. Not sure about removing original Daniel's mines.

SiegeSpud:
@Mailaender Noted. Metadata will be baked into sprites from now on.

@Dzierzan I wanted to retain the current randomnes of minefields in a more unified way. Regarding Daniel's sprites (from the left): red dots are to simple, rectangles look more like concrete slabs, triangles could pass as mines, spheres look similar to current water mines. I just added a little remapable green band. I'd like to reuse the rectangle and triangle mines as tank traps in the future with some minor modifications and shadows.

Dzierzan:
No, I meant keep original mines are a seperate png file. It might find another use.

Mailaender:
I will probably re-use it for scrap.

PR #789, opened 2023-09-15
Add air-transport squad / repair-squad for AI

dnqbob:
Main stuffs:

1. `IssueOrderToBot` and `ExternalBotOrdersManager`
The idea comes from `AutoDeployer` but more lightweight and less hardcoded, and has no problem on replay.

2. Update `SendUnitToAttackBotModule`.
-- Add air transport combat.
-- Bomber pod have better logic, they will attack the ambusher and be back if target is already destroyed.
-- We can introduce repair squad that repairs low HP unit in base.

3. Update `LoadCargoBotModule`. AI players will load unit for each other which should be more funs.

Other stuffs:
1.  Add AI of different difficulties, mainly to compare if the new changes I add make AI stronger or at least makes the game more funs.

2. Some fixes to AI and other stuff.


mods/hv/rules/bots.yaml
```diff
@@ -1,20 +1,39 @@
...
+		Condition: enable-easy-ai
+		Bots: easy
+	GrantConditionOnBotOwner@NormalAI:
+		Condition: enable-normal-ai
+		Bots: normal
+	GrantConditionOnBotOwner@HardAI:
+		Condition: enable-hard-ai
+		Bots: hard
+
+	CashTrickler@HardAI:
```
Mailaender, line 23: I don't want cheating AI.

mods/hv/rules/bots.yaml
```diff
@@ -1,20 +1,39 @@
 # License: CC-BY-SA-4.0
 
 Player:
-	ModularBot@RogueAI:
```
Mailaender, line 4: It is just one AI on purpose for easier maintenance and because they are in fact all the same.

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
```diff
@@ -0,0 +1,65 @@
...
+			entries.Add(new(issuer, order, chance));
+		}
+
+		void IBotTick.BotTick(IBot bot)
+		{
+			// "ManagerRunning = true" means IBotTick is running, and the game is
+			// 1. not a replay
+			// 2. not saved game still loading
+			// 3. the game running on the host where AI is enabled
+			ManagerRunning = true;
```
Mailaender, line 49: Why are you not using `world.IsReplay` then?

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
```diff
@@ -0,0 +1,65 @@
...
+
+		public void AddEntry(Actor issuer, string order, int chance)
+		{
+			entries.Add(new(issuer, order, chance));
+		}
+
+		void IBotTick.BotTick(IBot bot)
+		{
+			// "ManagerRunning = true" means IBotTick is running, and the game is
+			// 1. not a replay
```
Mailaender, line 46: Why are you not using `world.IsReplay` then?

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
```diff
@@ -0,0 +1,65 @@
...
+	{
+		public override object Create(ActorInitializer init) { return new ExternalBotOrdersManager(init.Self, this); }
+	}
+
+	public class ExternalBotOrdersManager : ConditionalTrait<ExternalBotOrdersManagerInfo>, IBotTick
+	{
+		readonly List<(Actor Actor, string Order, int Chance)> entries = new();
+		readonly World world;
+
+		public bool ManagerRunning { get; private set; }
```
Mailaender, line 29: Why are you hard-coding trait fields here?

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
```diff
@@ -0,0 +1,65 @@
...
+ */
+#endregion
+
+using System.Collections.Generic;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.HV.Traits
+{
+	[Desc("Allows the player to issue the orders from actors with " + nameof(IssueOrderToBot) + ".")]
```
Mailaender, line 18: This seems to be conceptually wrong. Each bot module should use `bot.QueueOrder` not another trait.

Mailaender:
Can you please not put everything into one huge pull request? Split this off. `No ShakeOnDeath for Landed Pod` is completely unrelated. The code changes seem very problematic, but `Increase aatank building rate for AI` and `BuildingDelays for tech building` seem like a good idea.

OpenRA.Mods.HV/Traits/BotModules/SendUnitToAttackBotModule.cs
```diff
@@ -74,112 +96,202 @@ public SendUnitToAttackBotModule(Actor self, SendUnitToAttackBotModuleInfo info)
...
-			minAssignRoleDelayTicks = world.LocalRandom.Next(0, Info.ScanTick);
+			// and we divide preparing stage, disguising stage and attacking stage for PERF.
+			prepareAttackTicks = world.LocalRandom.Next(0, Info.ScanTick);
+			assignAttackTicks = prepareAttackTicks + Info.ScanTick / 2;
 		}
 
 		void IBotTick.BotTick(IBot bot)
 		{
-			if (--minAssignRoleDelayTicks <= 0)
+			// TODO: if Open-HV is going to have disguising unit which needs AI support,
```
Mailaender, line 115: That is not planned.

OpenRA.Mods.HV/Traits/BotModules/SendUnitToAttackBotModule.cs
```diff
@@ -74,112 +96,202 @@ public SendUnitToAttackBotModule(Actor self, SendUnitToAttackBotModuleInfo info)
...
+			unitCannotBeOrdered = a => isInvalidActor(a) || a.Owner != player;
+			unitCannotBeOrderedOrIsBusy = a => unitCannotBeOrdered(a) || !(a.IsIdle || a.CurrentActivity is FlyIdle);
 			desireIncreased = 0;
 		}
 
 		protected override void TraitEnabled(Actor self)
 		{
 			// Avoid all AIs reevaluating assignments on the same tick, randomize their initial evaluation delay.
-			minAssignRoleDelayTicks = world.LocalRandom.Next(0, Info.ScanTick);
+			// and we divide preparing stage, disguising stage and attacking stage for PERF.
```
Mailaender, line 108: There is no disguise in OpenHV.

mods/hv/rules/bots.yaml
```diff
@@ -1,20 +1,39 @@
...
-	BaseBotModule@RogueAI:
-		RequiresCondition: enable-rogue-ai
-	BuilderBotModule@RogueAI:
-		RequiresCondition: enable-rogue-ai
+		Type: normal
+	ModularBot@HardAI:
+		Name: Rampage AI
+		Type: hard
+	GrantConditionOnBotOwner@EasyAI:
+		Condition: enable-easy-ai
```
Mailaender, line 14: I doubt this one will be easier, just less capable.

Mailaender (changes requested):
Please avoid hacks and submit this in smaller pieces.

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
dnqbob: because `BotTick is running can cover all three conditions I list here, and I am not sure we can use `world.IsReplay` to check the replay for loading game.

mods/hv/rules/bots.yaml
dnqbob: Sure, then I will delete those AI

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
dnqbob: If we cannot check loading saved game, `IssueOrderToBot` will add order to list endlessly while loading the game, which will result in ill behaviours when loading is done.

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
dnqbob: I guess you means `ManagerRunning`, `ManagerRunning` is ensure the bot tick is running and stop `IssueOrderToBot` add orders while it cannot consume.

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
dnqbob: The idea is: 
1. When actors with `IssueOrderToBot` have certain condition, it can use `IssueOrderToBot` to require bot player to give it an order.
2. `ExternalBotOrdersManager` recieve the requirement and give it an order.

It is a cheap way (compared to search the actors in world and give orders) to perform control in detail, and very flexible.

dnqbob: Idk if you agree with my explanation above. I think there maybe some misunderstandings in the description, but I am unable to think of better one on this.

OpenRA.Mods.HV/Traits/BotModules/SendUnitToAttackBotModule.cs
dnqbob: Then I will also remove the one in `AttackOptions`.

mods/hv/rules/bots.yaml
dnqbob: Without cheat AI, we will need AI's base management and strategy planning closed to human, or it will be easy to be defeated.

Current AI doesn't reach this level.

dnqbob: Anyway, if you don't need different AI then it can be deleted.

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
dnqbob: I find we can use `World.IsLoadingGameSave`, but I stick to just check if `BotTick` running, which is more simple and understandable.

dnqbob:
Remove unrelated commits and different AI are removed.
`SendUnitBotModule` remove the disguise related.

The remaining commits are related. `Increase aatank building rate for AI` and `BuildingDelays for tech building` is to make bot player not die too fast in AI test, or you will have to restart the game for several times to see the AI change.

Mailaender:
I meant you should submit them in their own easier to digest pull request not to ditch them.

dnqbob:
I will PR them later afte this is merged. Some commits like different AI focus on comparison those new bot module, they can be dropped and easy to re-write. 

I thought this PR was huge so I made it focus on bot module related.

dnqbob:
In addition, the "enable air-trans for AI" requires all changes in those bot module, only except the "LoadCargoBotModule up-to-date". I quite don't want to split air-transpot squad into a PR of "has dependcies" for hard to maintain when I need to change the bot module.

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
dnqbob: I found some interesting stuff in "ModularBot.cs":
1. Bot tick check if it should be actived only on host, at line 63 comment.
2. Bot tick is only actived when the game is not replay, at line 64.
3. Bot tick need the game is not `World.IsLoadingGameSave`, at line 86.

Therefore, my method here to check if bot tick is running is correct. We don't need any other checks, unless we move the checks to `IssueOrderToBot` and remove `IsManagerRunning`. Again, I don't think that is a good idea on both code maintaining and perf.

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
dnqbob: If we don't add the trait field `ManagerRunning`, then we need to add  `Game.IsHost || World.IsReplay || World.IsLoadingGameSave` to replace `ManagerRunning` in `IssueOrderToBot`. 

I don't think that is a good idea, for there is no trait on regular actor should care about those fields, which looks like they may cause desync. `ManagerRunning` is more acceptable in this case.

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
Mailaender: I mean the actors should not request to be ordered around. The bot player should both decide what action to do and order it. See also this change https://github.com/OpenHV/OpenHV/commit/7f8833febb23e98bc6ef6569bf7d0d47c77d76eb where I removed a similar hack.

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
Mailaender: You are relying on `ModularBot` to behave this way and hope that it stays. It is better to directly check against `IsLoadingGameSave` and `IsReplay` instead. Also keeps the code readable by itself.

mods/hv/rules/bots.yaml
```diff
@@ -70,18 +70,25 @@ Player:
 			uplink: 1
 			field: 1
 			silo: 1
+		BuildingDelays:
+			radar: 4000
+			radar2: 4000
+			howitzer: 8000
+			field: 8000
+			silo: 9000
+			uplink: 9000
```
Mailaender, line 79: This is a good idea. Can you submit it separately?

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
Mailaender: Another reason to get rid of this trait.

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
dnqbob: I mean all `IBot.BotTick` only runs by `ModularBot`, nowhere else.  So `BotTick` is running <=> `ModularBot.ITick` is fully functional <=> The game is not host, not replay and not in loading. From logic it is correct.

dnqbob: Even if we need to add `Game.IsHost || World.IsReplay || World.IsLoadingGameSave`, it should never be put inside the `IBot.BotTick`, it is superfluous.

dnqbob: Plus, if let us assume my thought that
> relying on `ModularBot` to behave this way

has flaws.

Then, yeah, we should add those checks to `IBot.BotTick` in `ExternalBotOrdersManager`, **and should not we add those checks to all other bot modules that use `IBot.BotTick`**? And why they don't even need that?

OpenRA.Mods.HV/Traits/BotModules/ExternalBotOrdersManager.cs
dnqbob: Btw, I also think https://github.com/OpenHV/OpenHV/commit/7f8833febb23e98bc6ef6569bf7d0d47c77d76eb is a good change but I prefer the reason that actors belong to AI should at least request order to hope the player to change itself, instead of directly changing itself by using activity. 

It is a matter of fair micro-manage of the order lantency in network. We all know that if the server is not good and order lantency >10 ticks, then actor change based on the order given will have a latency of >10 ticks.  

However, if actor change itself by using activity directly, there is no latency. This will be very unfair to players if they play against AI in mutiplayer.

mods/hv/rules/bots.yaml
Mailaender: Please rebase onto https://github.com/OpenHV/OpenHV/pull/800.

Mailaender:
Thanks for your submission, but I don't want units ordering themselves around using a bot module relay.

PR #791, opened 2023-09-15
Replace land mines

SiegeSpud:
Replaced land mines with updated ones based on Daniel Cook's artwork.


mods/hv/bits/mines.yaml
```diff
@@ -1,5 +1,5 @@
-FrameSize: 20,20
-FrameAmount: 4
-Offset: 0,0
-Author: Daniel Cook
-License: CC-BY-3.0-US
```
Mailaender, line 5: Looks like you introduced a line ending problem.

Mailaender:
Again, you didn't bake in the metadata.

Dzierzan:
Besides, they look so custom, you could easily remove Daniel from credits.

Mailaender:
They are based on the 4th sprite.

PR #800, opened 2023-10-02
BuildingDelays for tech building, it benefits AI on testing and gameplay

dnqbob:


Mailaender:
This avoids the AI to air rush unreasonably often.

PR #812, opened 2023-10-08
Set explored map checkbox to be enabled by default.

Dzierzan:


Mailaender:
This will reveal the enemy starting position:

image

Dzierzan:
Ughhh, that's unfortune. And kinda shows a problem with that lighting for Base.

Mailaender:
I just fixed it https://github.com/OpenHV/OpenHV/commit/5f445b223e924992f4048707fd9dded6dd47f640 so we can use this for multiplayer.

PR #813, opened 2023-10-08
Added transformation sounds for miner actor.

Dzierzan:
However de-transformation sound is played at the end of process, most likely OpenRA bug.


Dzierzan:
Apparently `Building` trait needed `UndeploySounds`. Now it works nicely.

PR #815, opened 2023-10-08
Added random twinkle effects for resources.

Dzierzan:
twinkle


Mailaender:
They are actually icons for underground resources, but it establishes a nice visual hint that the play should do something with it because it is important.

PR #820, opened 2023-10-13
Update status on Fire Watch when fires are clear

JovialFeline:
Fixes an issue where the player can be left with no more fires, but also no official victory because the burn percentage has gone unchanged.


Mailaender:
But then we don't get failed objectives while playing.

JovialFeline:
That's not what testing showed for me, but "Add status update when fires are clear" may have been more precise. The condition for an update is just shifted from

> The percentage of burned forest has changed

to

> The percentage of burned forest has changed **OR** all trees are done burning (either by being saved or transforming to ash)?

It's possible on main to extinguish the last few fires without a percentage change and ensure future updates are skipped, which is why I added the second bit.

PR #825, opened 2023-10-14
Increase Storage HP

### LINKED ISSUE #821: Storages could have more HP
It should be interesting to have more HP on Storages, since it's easy to focus attack on it and kill the player's resource income with ease.

[Storage HP Buff Idea - hv-2023-10-13T202237Z.orarep.zip](https://github.com/OpenHV/OpenHV/files/12898688/Storage.HP.Buff.Idea.-.hv-2023-10-13T202237Z.orarep.zip)

WilliamWeberBerrutti:
This PR attempts to fix #821

### DIFF
--- a/mods/hv/rules/buildings.yaml
+++ b/mods/hv/rules/buildings.yaml
@@ -1585,7 +1585,7 @@ STORAGE:
 		Order: 10
 		Category: Buildings
 	Health:
-		HP: 55000
+		HP: 125000
 	Armor:
 		Type: Steel
 	RevealsShroud:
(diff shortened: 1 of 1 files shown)


mods/hv/rules/buildings.yaml
```diff
@@ -1585,7 +1585,7 @@ STORAGE:
 		Order: 10
 		Category: Buildings
 	Health:
-		HP: 55000
+		HP: 125000
```
Dzierzan, line 1588: Make it 100000.

Dzierzan (changes requested):

Dzierzan:
Squash commits.

PR #826, opened 2023-10-14
Balance military unit production power costs

### LINKED ISSUE #824: Air is too dominant
Air is too dominant during matches: the Starports cost too low power compared to other military production buildings.

What about changing Starport power requirement to 30, Vehicle Factory to 120 and Harbor to 90?

WilliamWeberBerrutti:
This PR attempts to fix #824.

### DIFF
--- a/mods/hv/rules/buildings.yaml
+++ b/mods/hv/rules/buildings.yaml
@@ -683,7 +683,7 @@ STARPORT:
 		PrimaryCondition: primary
 		SelectionNotification: PrimaryBuildingSelected
 	Power:
-		Amount: -10
+		Amount: -120
 	Targetable:
 		TargetTypes: Ground, Structure
 	WithTextDecoration@PRIMARY:
@@ -1795,7 +1795,7 @@ HARBOR:
 	ProductionBar:
 		ProductionType: Ship
 	Power:
-		Amount: -150
+		Amount: -120
 	ProvidesPrerequisite@Generic:
 		Prerequisite: harbor
 		RequiresCondition: !powerdown


mods/hv/rules/buildings.yaml
```diff
@@ -683,7 +683,7 @@ STARPORT:
 		PrimaryCondition: primary
 		SelectionNotification: PrimaryBuildingSelected
 	Power:
-		Amount: -10
+		Amount: -30
```
Dzierzan, line 686: Make it `-120`.

mods/hv/rules/buildings.yaml
```diff
@@ -752,7 +752,7 @@ FACTORY3:
 	RevealsShroud@Hacked:
 		Range: 2c0
 	Power:
-		Amount: -150
+		Amount: -120
```
Dzierzan, line 755: Keep it `-150`.

mods/hv/rules/buildings.yaml
```diff
@@ -1795,7 +1795,7 @@ HARBOR:
 	ProductionBar:
 		ProductionType: Ship
 	Power:
-		Amount: -150
+		Amount: -90
```
Dzierzan, line 1798: Make it `-120`.

Dzierzan (changes requested):

mods/hv/rules/buildings.yaml
Mailaender: Maybe even more. The `-10` is definitely a mistake as you only need one factory, not multiple air fields.

Dzierzan:
LGTM, squash commits and make it ready for merging.

PR #828, opened 2023-10-14
Balance anti air tanks

### LINKED ISSUE #827: Anti Air Tanks Imbalance
It seems Yuruki's AA Tank is way stronger than Synapol's AA Tank.

WilliamWeberBerrutti:
This PR attempts to fix #827.


Rampoina:
We tested 20 gunships vs 20 AA tanks from each faction until we reached similar results (6-11 planes surviving).
Without the changes the lighting AA tanks were very good and Yuruki's very weak in comparison.

PR #833, opened 2023-10-15
Filled top right corner of sidebar art.

Dzierzan:


Mailaender:
I see the difference, but you need to explain to me why this change.

Dzierzan:
It was annoying seeing changing pixels in the top right corner when you moved around.

PR #836, opened 2023-10-16
Updated `Doubles` map.

Dzierzan:


mods/hv/maps/doubles/map.yaml
```diff
@@ -399,3 +368,241 @@ Actors:
...
+	Actor159: dark_light
+		Owner: Neutral
+		Location: 105,49
+	Actor160: island_light
+		Owner: Neutral
+		Location: 54,26
+	Actor161: seamnstr
+		Owner: Neutral
+		Facing: 384
+		Location: 54,28
```
Mailaender, line 467: image
is unable to move

PR #841, opened 2023-10-18
Updated `Mount Brosnan` map.

Dzierzan:


Mailaender:
I don't like the flat bridges.

PR #843, opened 2023-10-19
Added shadow layer for bridges.

Dzierzan:
Testcase:

image


Mailaender:
I tried to convert the fake actors into a proper sprite overlay
https://github.com/Mailaender/OpenHV/compare/bridge-shadow-layer
but there are problems with flickering and overlap.

PR #847, opened 2023-10-22
Updated description for Copter, Plane and Elite Pod.

Dzierzan:


mods/hv/rules/aircraft.yaml
```diff
@@ -9,7 +9,7 @@ PLANE1:
 		Queue: Aircraft
 		BuildPaletteOrder: 30
 		Prerequisites: ~starport.yi
-		Description: Attack Ship armed with\na large chain gun.\n  Strong vs Pods and Buildings\n  Weak vs Tanks
+		Description: Attack Ship armed with\na large chain gun.\n  Strong vs Pods, Buildings and Aircrafts\n  Weak vs Tanks
```
Mailaender, line 12:
```suggestion
		Description: Attack Ship armed with\na large chain gun.\n  Strong vs Pods, Buildings and Aircraft\n  Weak vs Tanks
```

https://en.wiktionary.org/wiki/aircraft#Noun

PR #853, opened 2023-10-22
Renamed `Tech Reconstructor` to `Nano Reconstructor`.

Dzierzan:


Mailaender:
All other tech buildings consist of two words.

Dzierzan:
As far as I remember, you were removing word `tech` from all captureable `tech` buildings. How come `two words` is a rule here?

Mailaender:
Extractor (what does it extract) → Resource Extractor is more explainable. I agree that "Tech Reconstructor" can be improved. I didn't come up with something better yet.

mods/hv/rules/buildings.yaml
```diff
@@ -2412,7 +2412,7 @@ RECONSTRUCTOR:
 	Selectable:
 		Bounds: 2048, 2457, 0, -204
 	Tooltip:
-		Name: Tech Reconstructor
+		Name: Reconstructor
```
Mailaender, line 2415: Nano Reconstructor
Universal Reconstructor

something like that

Mailaender: Reconstructor Hub/Facility/Center?

Dzierzan: Nano Reconstructor sounds OK.

PR #862, opened 2023-11-02
Balanced Ramp Buggy and Gatling Bike.

Dzierzan:
- HP decreased by 500 for both vehicles.
- Increased damage against pods.
- Gatling Bike's speed was increased.

@WilliamWeberBerrutti 

Please test it and see if it's any better. The reason why I decreased HP is because I don't want those vehicles to become way too powerful. Pods need to stand a chance too.


WilliamWeberBerrutti:
@Rampoina and me playtested and we concluded these changes need to be made. It's not in an ideal state, but better than before.

Screenshot_20231104_181343

PR #888, opened 2023-12-26
Balance Air Units

WilliamWeberBerrutti:
This PR attemps to fix https://github.com/OpenHV/OpenHV/issues/882 and fix https://github.com/OpenHV/OpenHV/issues/906, rebalances Banshees and Speeders, rebalances air transports of both factions and rebalances anti-air turrets.

### DIFF
--- a/mods/hv/rules/aircraft.yaml
+++ b/mods/hv/rules/aircraft.yaml
@@ -26,10 +26,14 @@ PLANE1:
 	RevealsShroud@Hacked:
 		Range: 4c0
 		Type: GroundPosition
-	Armament@Primary:
-		Weapon: AircraftChainGun
-		LocalOffset: 256,0,0
+	Armament@Ground:
+		Weapon: AircraftChainGunGround
 		MuzzleSequence: muzzle
+		LocalOffset: 50,0,0
+	Armament@Air:
+		Weapon: AircraftChainGunAir
+		MuzzleSequence: muzzle
+		LocalOffset: 50,0,0
 	AttackAircraft:
 		FacingTolerance: 20
 		PersistentTargeting: false
@@ -162,10 +166,14 @@ COPTER:
 		Offset: 200,0,-100
 	AttackTurreted:
 		RequiresCondition: airborne
-	Armament:
-		Weapon: HelicopterChainGun
-		LocalOffset: 50,0,0
+	Armament@Ground:
+		Weapon: HelicopterChainGunGround
 		MuzzleSequence: muzzle
+		LocalOffset: 50,0,0
+	Armament@Air:
+		Weapon: HelicopterChainGunAir
+		MuzzleSequence: muzzle
+		LocalOffset: 50,0,0
 	Selectable:
 		Bounds: 1228, 1228
 	WithMuzzleOverlay:
--- a/mods/hv/rules/aircraft.yaml
+++ b/mods/hv/rules/aircraft.yaml
@@ -35,7 +35,7 @@ PLANE1:
 		MuzzleSequence: muzzle
 		LocalOffset: 50,0,0
 	AttackAircraft:
-		FacingTolerance: 20
+		FacingTolerance: 92
 		PersistentTargeting: false
 		OpportunityFire: false
 	Aircraft:
@@ -103,7 +103,7 @@ PLANE2:
 		Delay: 100
 		Count: 1
 	AttackAircraft:
-		FacingTolerance: 20
+		FacingTolerance: 92
 		PersistentTargeting: false
 		OpportunityFire: false
 	Aircraft:
--- a/mods/hv/rules/aircraft.yaml
+++ b/mods/hv/rules/aircraft.yaml
@@ -311,7 +311,7 @@ COPTER2:
 	Selectable:
 		DecorationBounds: 2048, 1843
 	Cargo:
-		Types: Scout, Vehicle
+		Types: Scout
 		MaxWeight: 8
 		AfterUnloadDelay: 40
 		RequiresCondition: !carrying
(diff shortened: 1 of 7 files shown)


Mailaender:
Needs a rebase. Also, can you summarize your changes for the changelog? Your commit messages are pretty bad.

WilliamWeberBerrutti:
Maybe it should be better if I could fixup some commits? During the development, I was so focused on playtesting and experimenting that I didn't invest much time on commit messages.

mods/hv/weapons/missiles.yaml
```diff
@@ -295,27 +288,28 @@ TowerMissile:
 DropshipMissile:
```
Mailaender, line 288: You removed `DropshipMissile`. Why is it still here?

mods/hv/rules/defaults.yaml
```diff
@@ -389,7 +389,6 @@
 		WaitDistanceFromResupplyBase: 4c0
 		TakeOffOnResupply: true
 		VTOL: true
-		LandableTerrainTypes: Clear, Ore, Crater, Road, Grass, Grass Pit, Mountain, Snow, Red Snow, Sand, Stone, Tech, Rail, Blocked, Dirt
```
Mailaender, line 392: Why are you removing this and duplicating it elsewhere?

mods/hv/weapons/missiles.yaml
```diff
@@ -295,27 +288,28 @@ TowerMissile:
...
 	Inherits: ^AntiGroundMissile
 	Inherits: BigSplash
-	ReloadDelay: 100
-	Range: 8c0
+	ReloadDelay: 70
+	Range: 6c0
+	MinRange: 0c768
```
Mailaender, line 293: And why did you make changes to it?

mods/hv/rules/aircraft.yaml
```diff
@@ -299,16 +310,8 @@ COPTER2:
...
-		TurnSpeed: 80
-		Offset: 600,0,100
-	AttackTurreted:
-		RequiresCondition: airborne
-	Armament:
-		Weapon: Vulcan
-		LocalOffset: 0,0,-400
 	Cargo:
-		Types: Scout
+		Types: Scout, Vehicle
```
Mailaender, line 314: This is reserved for infantry.

mods/hv/weapons/missiles.yaml
WilliamWeberBerrutti: Since Dropships won't have weapons anymore, their weapon actors should be removed.

mods/hv/rules/defaults.yaml
WilliamWeberBerrutti: So I don't have to remove it from Elite Pods, Attack Helicopters, Banshees. By default, maybe it should be better to not have 'LandableTerrainTypes' in the base actor, and have this property on actors that should be able to land, but I may be wrong.

mods/hv/rules/aircraft.yaml
```diff
@@ -402,7 +406,6 @@ BALLOON:
 DROPSHIP:
 	Inherits: ^Helicopter
 	Inherits@Smoke: ^SmokeEmitter
-	Inherits@AutoTarget: ^AutoTargetGroundAssaultMove
```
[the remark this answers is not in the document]
WilliamWeberBerrutti: Even if Dropships do not attack anymore? I'll add it back then!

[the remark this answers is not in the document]
WilliamWeberBerrutti: It seems it's not breaking. Maybe I shouldn't add it.

mods/hv/weapons/missiles.yaml
WilliamWeberBerrutti: Removed now!

mods/hv/rules/aircraft.yaml
```diff
@@ -411,11 +414,13 @@ DROPSHIP:
 		Queue: Aircraft
 		BuildPaletteOrder: 50
 		Prerequisites: ~starport.yi
-		Description: Vehicle Transport Shuttle.\n  Armed with light missiles.\n  Can load pods\n  and lift one vehicle.
+		Description: Vehicle Transport Shuttle.\n  Armed with light machine gun.\n  Can load pods\n  and lift one vehicle.
```
Mailaender, line 417: Why do you say this has machine guns when you only removed the missiles?

mods/hv/rules/aircraft.yaml
```diff
@@ -272,11 +280,13 @@ COPTER2:
 		Queue: Aircraft
 		BuildPaletteOrder: 50
 		Prerequisites: ~starport.sc
-		Description: Vehicle Transport Helicopter.\n  Armed with light machine gun.\n\n  Can load pods\n  and lift one vehicle.
+		Description: Vehicle Transport Helicopter.\n  Armed with light missiles.\n  Can load pods\n  and lift one vehicle.
```
Mailaender, line 283: You just removed the machine gun and forgot about the missiles.

Mailaender:
Can you give a reason why you removed all weapons from the transport aircraft?

WilliamWeberBerrutti:
> Can you give a reason why you removed all weapons from the transport aircraft?

It's because air transports purpose is to transport units, not attack, so I increased their resistance and made them having Heavy armor type (similarly to what happens with Transport Tank and Water Transport).

When it's added an anti-air air unit, this unit could be good to beat air units, including air transports. I know having a mass of helis and planes

Maybe it could be interesting to let they attack if they are loaded by pods, like it happens with Transport Tanks? Maybe this could be done for Water Transports too?

PR #913, opened 2024-01-13
Balance water

WilliamWeberBerrutti:
This PR attempts to contemplate discussion https://github.com/OpenHV/OpenHV/discussions/910 and fix https://github.com/OpenHV/OpenHV/issues/919 and increases boats LOS.

### DIFF
--- a/mods/hv/sequences/ships.yaml
+++ b/mods/hv/sequences/ships.yaml
@@ -29,6 +29,18 @@ boat2:
 		Filename: bits/units-icons.png
 		Start: 16
 
+boat4:
+	idle:
+		Filename: bits/boat2.png
+		Facings: 8
+	turret:
+		Filename: bits/boat2.png
+		Start: 8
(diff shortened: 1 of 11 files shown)


Mailaender:
Needs a rebase.

mods/hv/sequences/ships.yaml
```diff
@@ -29,6 +29,18 @@ boat2:
 		Filename: bits/units-icons.png
 		Start: 16
 
+boat4:
+	idle:
+		Filename: bits/boat2.png
```
Mailaender, line 34: I think this will only lead to confusion if different boats look the same.

WilliamWeberBerrutti: Oh yea I forgot to ask to have new sprites for the Electric Boat, if you consider both factions should have Railgun / Electric Boats, the same way they have for tanks.

PR #941, opened 2024-03-17
Revamp Orbital Strike to look more like such.

GraionDilach:
https://github.com/OpenHV/OpenHV/assets/1136302/321493ba-f942-4410-a600-6c1d7426060b

https://github.com/OpenHV/OpenHV/assets/1136302/d5f7979e-ab63-4c9d-b5f0-13538136d6d0

Still needs some cleanup, like trimming the unused frames out of the artwork and maybe it coulda get a pixel trailer animation of some sort as well.

Actual players might also discuss/tweak the projectile speed and the activation delays. Note that the damage/warhead size were unchanged.


Dzierzan:
It looks quite nice.

mods/hv/rules/buildings.yaml
```diff
@@ -1581,13 +1588,9 @@ UPLINK:
...
 		SelectTargetTextNotification: notification-select-target
 		IncomingSpeechNotification: OrbitalRailgunFired
-	Power:
-		Amount: -150
-	MustBeDestroyed:
-		RequiredForShortGame: false
-	WithIdleOverlay@Lights:
-		Sequence: animation
-		RequiresCondition: !build-incomplete
+		ActivationDelay: 7
```
GraionDilach, line 1591: Rebased this out to 0, because the description says instant.

GraionDilach:
Should I delete out the now unused code (RailgunPower) and trim out the unused framesw from the railgun SFX?

WilliamWeberBerrutti:
What if it fired multiple projectiles in sequence instead of all at the same time?

GraionDilach:
> What if it fired multiple projectiles in sequence instead of all at the same time?

Would need a complete rebalancing on the weapon which I didn't wanted to do.

> What if it was a continuous beam, which had periodic radius damage for, maybe, 5 seconds?

Such an beam would need to be drawn using engine effects, which would look really jarring on a high scale next to the pixel effects - keep in mind that most engine lasers/electrics are thin to make them less jarring. This shouldn't be done.

Rusted Warfare looks wrong exactly for the same reason due to blatantly mixing pixelart with engine-effect particles and 32bit explosions.

OcelotWalrus:
> What if it fired multiple projectiles in sequence instead of all at the same time?

This seems to be a pretty good idea.

GraionDilach:
Why?

What would a pattern add here? Also, what would such a pattern look like? A randomized pattern doesn't look something the project would accept (considering that even the artillery support power accurately hits the target cell), so only fixed patterns would be acceptable. Combining non-radius patterns with the epicenter-based damage delivery break immersion and shifting from the epicenter-circular damage model to a complete pattern-based damage model should be out of scope of this PR, considering that this was supposed to be a visual polish.

OcelotWalrus:
It'd be a fixed pattern, making so the different projectiles don't fire exactly at the same time would make it look way better in my opinion.

GraionDilach:
> It'd be a fixed pattern, making so the different projectiles don't fire exactly at the same time would make it look way better in my opinion. Just a visual change.

Either this means there will be laser explosions without damage delivery happening in their triggering frames (which would be bad because it defies player expectations, which are already problematic, and something this proposal tries to improve, see footnote), either there will be actually tied damage delivery to the later explosions in which case this wouldn't be a visual change, and has actual balance implications.

For the sake of discussion, the patterns reasonable within the current internal consistency rules of the project are either the forward or the reverse cones, where there would be multiple waves of circular patterns fired with either increasing or decreasing radius - meaning either a consistently growing or shrinking spread rate with each wave. Such a redesign wouldn't "just be a visual change".

Reminder, the circular pattern in this PR is introduced only because there's no better option to visually represent the actual spread of the weapon using the current artwork. The warhead damage data is unchanged in this PR and was carried over from the current stage. There's a gameplay issue with the current implementation that the spread is a lot bigger than what the art implies.

Mailaender (approved):

PR #957, opened 2024-05-07
Add InstantHitWithFakeBullets

dnqbob:
example

Spawns several fake bullets with interval (or set interval to 0 if you want a shot gun effect) upon each fire.

### DIFF
--- a/OpenRA.Mods.HV/Projectiles/InstantHitWithTracers.cs
+++ b/OpenRA.Mods.HV/Projectiles/InstantHitWithTracers.cs
@@ -1,0 +1,315 @@
+#region Copyright & License Information
+/*
+ * Copyright 2024 The OpenHV Developers (see CREDITS)
+ * This file is part of OpenHV, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System;
+using System.Collections.Generic;
+using OpenRA.GameRules;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common;
+using OpenRA.Mods.Common.Effects;
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Primitives;
+using OpenRA.Traits;
+using Util = OpenRA.Mods.Common.Util;
+
+namespace OpenRA.Mods.HV.Projectiles
+{
+	[Desc("Instant and usually direct-on-target projectile, with tracer ammunition.")]
+	public class InstantHitWithTracersInfo : IProjectileInfo
+	{
+		[Desc("The maximum/constant/incremental inaccuracy used in conjunction with the InaccuracyType property.")]
+		public readonly WDist Inaccuracy = WDist.Zero;
+
+		[Desc("Controls the way inaccuracy is calculated. Possible values are 'Maximum' - scale from 0 to max with range, 'PerCellIncrement' - scale from 0 with range and 'Absolute' - use set value regardless of range.")]
+		public readonly InaccuracyType InaccuracyType = InaccuracyType.Maximum;
+
+		[Desc("Projectile can be blocked.")]
+		public readonly bool Blockable = false;
+
+		[Desc("The width of the projectile.")]
+		public readonly WDist Width = new(1);
+
+		[Desc("Scan radius for actors with projectile-blocking trait. If set to a negative value (default), it will automatically scale",
+			"to the blocker with the largest health shape. Only set custom values if you know what you're doing.")]
+		public readonly WDist BlockerScanRadius = new(-1);
+
+		[Desc("Amount of Tracers.")]
+		public readonly int TracerAmount = 0;
+
+		[Desc("Tracer spawn interval. Only useful when Amount > 0")]
+		public readonly int TracerSpawnInterval = 3;
+
+		[Desc("Tracer speed")]
+		public readonly int TracerSpeed = 1024;
+
+		[Desc("Tracer inaccuracy, use set value regardless of range")]
+		public readonly WDist TracerInaccuracy = WDist.Zero;
+
+		[Desc("Image to display.")]
+		public readonly string Image = null;
+
+		[SequenceReference(nameof(Image), allowNullImage: true)]
+		[Desc("Loop sequence of Image from this list while this projectile is moving.")]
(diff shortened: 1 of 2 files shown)


OpenRA.Mods.HV/Projectiles/InstantHitWIthFakeBullets.cs
```diff
@@ -0,0 +1,316 @@
...
+using OpenRA.Graphics;
+using OpenRA.Mods.Common;
+using OpenRA.Mods.Common.Effects;
+using OpenRA.Mods.Common.Graphics;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Primitives;
+using OpenRA.Traits;
+using Util = OpenRA.Mods.Common.Util;
+
+namespace OpenRA.Mods.SP.Projectiles
```
Mailaender, line 24:
```suggestion
namespace OpenRA.Mods.HV.Projectiles
```

OpenRA.Mods.HV/Projectiles/InstantHitWIthFakeBullets.cs
```diff
@@ -0,0 +1,316 @@
...
+{
+	[Desc("Instant and usually direct-on-target projectile, with traces effect like 3rd generation of CNC.")]
+	public class InstantHitWIthFakeBulletsInfo : IProjectileInfo
```
Mailaender, line 27:
```suggestion
	public class InstantHitWithFakeBulletsInfo : IProjectileInfo
```

OpenRA.Mods.HV/Projectiles/InstantHitWIthFakeBullets.cs
```diff
@@ -0,0 +1,316 @@
...
+			public int OverallTime;
+			public ContrailRenderable Contrail;
+			public Animation Anim;
+			public WPos EndPos;
+			public WPos Pos;
+			public WPos SourcePos;
+			public WAngle Facing;
+		}
+
+		public InstantHitWIthFakeBullets(InstantHitWIthFakeBulletsInfo info, ProjectileArgs args)
```
Mailaender, line 149:
```suggestion
		public InstantHitWithFakeBullets(InstantHitWIthFakeBulletsInfo info, ProjectileArgs args)
```

OcelotWalrus:
I don't see why that'd actually be useful.

dnqbob:
> have a weird looking shot angle

The problem is the fire muzzle sprites have initial offset which is conflicted with the "LocalOffsets" in `Armament`.
example
When set "LocalOffset: 900,0,0", the bullet trace is correct but the muzzle is wrong.

OcelotWalrus:
I still don't find a really useful case for that new trait.

GraionDilach:
This looks appropriate. I'd say drop the muzzle effect from the weapons using this trait though, they no longer need it (another option might be a really small circular muzzle anim akin to most RA2 gun muzzles to prevent exposing the sprite's unability to track the facing along but considering the size required for this I'd guess the size would be too small to be meaningful).

> I still don't find a really useful use for that new trait.

I can say the same about all your gameplay proposals so far.

GraionDilach:
TBH, if you combine ammo and GrantConditionOnAttack conditions on the decoration armament, you _can_ keep the sync, but that is a massive hack (and depending on the amount of possible reloaddelaymodifier condition sources could be way too tedious/fragile), so this is indeed a more reliable option, all things considered.

Regarding the technical implementation, I'm not quite certain if FakeBulletWrapper should remain as a struct or it should be promoted to a separate IEffect subclass that InstantHitWithFakeBullets just spawns and forgets thenafter.

OcelotWalrus:
> I can say the same about all your gameplay proposals so far.

If you think like that, then you could probably say that pretty much every pod or vehicle variant is not very useful.

dnqbob:
> TBH, if you combine ammo and GrantConditionOnAttack conditions on the decoration armament, you can keep the sync, but that is a massive hack (and depending on the amount of possible reloaddelaymodifier condition sources could be way too tedious/fragile), so this is indeed a more reliable option, all things considered.

Yep, some painful experience, then I finally decide to make this projectile and get rid of all the hacks.

And also some mods are not developed by 1 person, and decoration weapon system is casecaded with actual weapon, which means that if the balance maker in your team modify the actual armament, it may break the behaviors of decoration armament, cause some mess.

dnqbob:
Maybe I am a person forcus too

dnqbob:
> I just prefer how bullets visual effects are currently.

dnqbob:
> I just prefer how bullets visual effects are currently.

Maybe I am a person focus too much on consistency, but IMO, it is better that to make all gun weapon instant hit or all use the `Bullet` projectile that has a delay on impact. It is not very good for me to have some machine gun instant hit and some machine gun has a delay on impact with a bullet trace, which will confuse player. 

In addition, machine gun with a delay on impact behaves terrible when anti-air, which makes helicopter in HV weaker than plane in gameplay, and you say they all use "machine gun", huh.

dnqbob:
> Regarding the technical implementation, I'm not quite certain if FakeBulletWrapper should remain as a struct or it should be promoted to a separate IEffect subclass that InstantHitWithFakeBullets just spawns and forgets thenafter.

dnqbob:
> Regarding the technical implementation, I'm not quite certain if FakeBulletWrapper should remain as a struct or it should be promoted to a separate IEffect subclass that InstantHitWithFakeBullets just spawns and forgets thenafter.

Yes, that is not a bad idea. Then I can spawn the effect and make the parent effect of `InstantHitWithFakeBullets` end normally, not like now I have to keep the `InstantHitWithFakeBullets` when sub effect is not finished.

GraionDilach:
Exactly. I'd say that this as-is just migrates the hackish solution from MiniYAML to C#. I also think upstream agrees with the assessment, considering that similar logic led me to design the AS particle system as-is, whose subset was merged back to upstream's FloatingSprite.

dnqbob:
If so I am going to make an upstream version

dnqbob:
> Regarding the technical implementation, I'm not quite certain if FakeBulletWrapper should remain as a struct or it should be promoted to a separate IEffect subclass that InstantHitWithFakeBullets just spawns and forgets thenafter.

On second thought, I find there is no other place to use the "FakeBulletTraces" effect, becuase the hit point of fake bullet depends on where the actual projectile explode, which can be blocked and even be countered. if the actual projectile cannot hit instantly, the situation will be complicated.

GraionDilach:
Lasers (with the laser functioning as targetlaser) come to my mind as another usecase.

Issue #924, opened 2024-01-30
Custom Alert Sound for Recon Tank


WilliamWeberBerrutti:
What if the Recon Tank emitted a custom alert when enemy units are in its detection range? Something like "Enemies approaching" or maybe "Enemy movement detected"?

OcelotWalrus:
The would be very interesting indeed. Maybe scout aircraft should have this functionality too.

OcelotWalrus:
Actually after checking, there's already something that does the job you're describing, but it only works on vehicles units: 

https://github.com/OpenHV/OpenHV/blob/main/mods/hv/rules/defaults.yaml#L163

PR #959, opened 2024-05-13
Add The `EnemyWatcher` Trait To Recon Units

OcelotWalrus:
This PR fix the issue described in #924, by adding the `EnemyWatcher` attribute to the Recon Tank, that allows him to send messages when it detects enemy or neutral units. I also gave it to the Synapol Scout Balloon and the Yuruki Scout Saucer.


OcelotWalrus:
My bad I didn't read the trait description enough.

PR #962, opened 2024-05-15
Enhance Bot Support Power Decisions

OcelotWalrus:
This PR tries to enhance a bit the bots' support power decisions:
- bots will now drop pods against enemy air and ground units, instead of just structures
- bots will now flush bombers against enemy ground units, instead of just structures
- bots will now launch howitzer attacks against enemy ground units, instead of just structures

### DIFF
--- a/mods/hv/rules/bots.yaml
+++ b/mods/hv/rules/bots.yaml
@@ -219,7 +219,7 @@ Player:
 				Consideration@1:
 					Against: Enemy
 					Types: Structure
-					Attractiveness: 1.5
+					Attractiveness: 2
 					TargetMetric: None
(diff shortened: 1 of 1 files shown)


mods/hv/rules/bots.yaml
```diff
@@ -228,24 +228,42 @@ Player:
 					Attractiveness: -5
 					TargetMetric: None
 					CheckRadius: 8c0
+				Consideration@3:
+					Against: Enemy
+					Types: Air, Ground
```
Mailaender, line 233: Throwing a few pods into the enemy main battle group is a pretty bad decision. They are also not effective against air units, but rather countered by them.

PR #965, opened 2024-05-22
Fix MakeFile Install Data

OcelotWalrus:
This RP adds a new line to the function that install the mod's data when running `sudo make install`, so it deletes the content of the directory before installing the mod. Before this, if you, for example, renamed a file in the mod data, and then run make install, the mod installation folder will contain the renamed file, and the original, which's a problem. This PR fixes this problem.

### DIFF
--- a/packaging/functions.sh
+++ b/packaging/functions.sh
@@ -86,5 +86,6 @@ install_data() (
 	cp -r "${SRC_PATH}/mods/common" "${DEST_PATH}/mods/"
 
 	echo "Installing Hard Vacuum mod files to ${DEST_PATH}"
+ 	rm -r "${DEST_PATH}/mods/hv/"*
 	cp -r "${SRC_PATH}/../mods/hv" "${DEST_PATH}/mods/"
 )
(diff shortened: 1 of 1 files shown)


GraionDilach:
I think this belongs to OpenRA/OpenRAModSDK.

packaging/functions.sh
```diff
@@ -86,5 +86,8 @@ install_data() (
 	cp -r "${SRC_PATH}/mods/common" "${DEST_PATH}/mods/"
 
 	echo "Installing Hard Vacuum mod files to ${DEST_PATH}"
+ 	if [ -d "${DEST_PATH}/mods/hv/" ]; then
+ 		rm -r "${DEST_PATH}/mods/hv/"*
```
Mailaender, line 90: This belongs to an `uninstall` target.

OcelotWalrus: ?

Mailaender: `make install` will create files and `make uninstall` is the counterpart that deletes them. https://www.gnu.org/software/make/manual/html_node/Standard-Targets.html Don't mix the two.

OcelotWalrus: It doesn't mix them. It just clears the content before installing. I don't see how that's a problem.

PR #969, opened 2024-05-23
New Mission: Harbor Defense

OcelotWalrus:
Adds a a new mission, and do some stuff to the missions panel. (part of closed #960)


mods/hv/missions.yaml
```diff
@@ -8,3 +8,4 @@ Yuruki Campaign:
 
 Synapol Campaign:
 	crater-defense
+	harbor-defense
```
Mailaender, line 11: When I did the Tower Defense mini-game, it was merely just for fun and to show the flexibility of what is possible. Let's not have a whole TD campaign.

OcelotWalrus: It's not a whole Tower Defense campaign, with #970, there's also a conquest campaign to the Synapol campaign.

PR #970, opened 2024-05-23
New Mission: Asteroid Invasion

OcelotWalrus:
Adds a a new mission, and do some stuff to the missions panel. (part of closed #960)

Be aware that this mission hasn't been fully tested yet.


mods/hv/maps/asteroid-invasion/artwork/buildings-icons.yaml
```diff
@@ -0,0 +1,5 @@
+FrameSize: 63,47
+FrameAmount: 39
```
Mailaender, line 2: Why are you duplicating a whole sprite sheet?

mods/hv/maps/asteroid-invasion/asteroid-invasion.lua
```diff
@@ -0,0 +1,111 @@
+--[[
+   Copyright 2021-2023 The OpenHV Developers (see AUTHORS)
```
Mailaender, line 2:
```suggestion
   Copyright 2024 The OpenHV Developers (see AUTHORS)
```

mods/hv/maps/asteroid-invasion/artwork/buildings-icons.yaml
OcelotWalrus: Right.

Mailaender:
The enemy does not properly attack the base:

grafik

Mailaender:
If you don't have a vehicle factory, most of the units like the main battle tank are hidden:

grafik

OcelotWalrus:
> The enemy does not properly attack the base:
> 
> grafik

It's not a bug: the yuruki forces just stays at the colony to defense against any incoming synapol attack. They detected some signs of movement where the player is, but they have no clue.

Mailaender:
The problem is that with initial trade platform before factory is that vehicles are invisible and not greyed out because in a regular match that build order is not possible.

Mailaender:
If you remove the factory prerequisites, you create inconsistent game rules compared to multiplayer matches. Either remove the trade platform or make the prerequisites visible.

OcelotWalrus:
Fixed everything you said!

OcelotWalrus:
The workflow failure actually happens because of the main branch from what I've seen.

mods/hv/maps/asteroid-invasion/map.yaml
```diff
@@ -0,0 +1,1426 @@
...
+		Owner: Synapol Corporation
+		Location: 37,10
+		Facing: 670
+	Actor152: balloon
+		Owner: Synapol Corporation
+		Facing: 384
+		Location: 30,4
+	Actor162: tradplat
+		Owner: Synapol Corporation
+		Location: 34,3
```
Mailaender, line 505: This is an impossible position:

grafik

which causes the units to exit through the cliff.

OcelotWalrus: Oopsie.

PR #984, opened 2024-06-13
Added artwork for Lightning Boat.

Dzierzan:
image

PR #985, opened 2024-06-16
Added Mothership unit.

Dzierzan:
Added super unit which is supposed to be used in single player campaign. The unit is right now overpowered but that's intended. Of course it is a subject to change if single player missions require that.

Depends on #984


OcelotWalrus:
Looks amazing!

OcelotWalrus (approved):

PR #992, opened 2024-06-18
Added missing voxel sources.

Dzierzan:


OcelotWalrus (commented):
`dropzoneplane` should be renamed to `airlifter`

OcelotWalrus (approved):

PR #996, opened 2024-06-22
Added missing bridge shadows for Harbour map.

Dzierzan:


Mailaender:
Thought this was on purpose.

PR #997, opened 2024-06-22
More consistent actor names

Mailaender:
as it got a little out of hand with the different schemes and the enumerations.


mods/hv/bits/submarine.yaml
Dzierzan, line 1: What happened here? Why differnet FrameSize and FrameAmount?

Mailaender: `submarine2` → `submarine`
`submarine? → `boomer`

Dzierzan: image

PR #998, opened 2024-06-22
Balance Mothership

WilliamWeberBerrutti:
This PR also rebalances many AA attacks against Heavy armored units, increases air transports HP and changes initial facings for all starport units.

For the interceptors to return, https://github.com/OpenHV/OpenHV/issues/1006 must be fixed.

**NOTE:** This PR ideally should be merged together with [Balance Battleship](https://github.com/OpenHV/OpenHV/pull/1096).

### DIFF
--- a/mods/hv/rules/aircraft.yaml
+++ b/mods/hv/rules/aircraft.yaml
@@ -285,7 +285,7 @@ CHOPPER:
 	UpdatesPlayerStatistics:
 		AddToArmyValue: true
 	Health:
-		HP: 30000
+		HP: 45000
 	Armor:
 		Type: Heavy
 	RevealsShroud:
@@ -419,7 +419,7 @@ DROPSHIP:
 	UpdatesPlayerStatistics:
 		AddToArmyValue: true
 	Health:
-		HP: 30000
+		HP: 45000
 	Armor:
 		Type: Heavy
 	RevealsShroud:
--- a/mods/hv/rules/aircraft.yaml
+++ b/mods/hv/rules/aircraft.yaml
@@ -547,7 +547,7 @@ DRONE2:
 	Tooltip:
 		Name: actor-drone2.name
 	Health:
-		HP: 5000
+		HP: 7500
 	Armor:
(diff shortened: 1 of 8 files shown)


mods/hv/rules/buildings.yaml
```diff
@@ -674,7 +674,7 @@ STARPORT:
 		RequiresCondition: !being-captured && !build-incomplete
 		SpawnOffset: 0,-256,0
 		ExitCell: 0,0
-		Facing: 896
+		Facing: 512
```
OcelotWalrus, line 677: Why changing that?

WilliamWeberBerrutti: This PR isn't ready to be reviewed yet, so anything can be changed.

mods/hv/rules/aircraft.yaml
```diff
@@ -682,20 +683,21 @@ AIRLIFTER:
...
 	Valued:
 		Cost: 6000
 	Tooltip:
 		Name: actor-mothership.name
 	Buildable:
 		Queue: Aircraft
 		BuildPaletteOrder: 100
-		Prerequisites: ~disabled
+		BuildLimit: 1
+		Prerequisites: radar, trader, techcenter
```
OcelotWalrus, line 695: You made it available to all factions.

OcelotWalrus: You must also add the `~starport.x` prerequisite or it'll still appear even if the player has no starport (if I'm right)

OcelotWalrus (approved):
The code looks good.

mods/hv/rules/buildings.yaml
Mailaender: That does not explain the change.

WilliamWeberBerrutti: Oh yea I forgot to answer this after removing the Draft! Pardon me!

I did this change so the Mothership (and consequently all units) face South, which seem to align better with the Starport shape.

PR #999, opened 2024-06-23
Check sprite metadata faster

Mailaender:
https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/


Makefile
```diff
@@ -270,7 +270,7 @@ endif
 
 check-bits: engine
 	@echo "Checking PNG sheet metadata..."
-	@$(MAKE) $(CHECK_TARGETS)
+	@$(MAKE) --jobs 4 $(CHECK_TARGETS)
```
GraionDilach, line 273: How about using nproc on this? Prior to the aforementioned newspost, the runners already had 3 threads.

Mailaender: On my local machine, this process on max threads can make the UI halt.

GraionDilach: ¯\\\_(ツ)\_/¯ Alright then.

PR #1000, opened 2024-06-23
Added Battleship unit.

Dzierzan:
And updated husk for Mothership and Battleship.

The only issue I have got so far is that turrets are gone when Battleship is destroyed. Yes I know that they're not added as husk currently does not inherit them. But I've got no idea how to KEEP the facing when Battleship is dead. @Mailaender any ideas?

Unless Husk trait is used? But I know it is used to keep the facing of the body, not sure if it also works for the turrets.


GraionDilach:
Husk only affects the body, it doesn't carry over the turret facings. It would need a separate `HuskTurret`(?) trait.

OcelotWalrus:
About the husk turret thing, I know that in D2K, vehicles husk keep the turret's facing.

Dzierzan:
No, they do not keep the facing. They use `ThrowsParticle` where a turret is a particle which is meant to rotate midair. But the time where it is in midair is so short, it might feel the facing is kept.

Dzierzan:
Well, I am thinking about simpler idea... the battleship could use use `ThrowsParticle`, so it would "eject" two turrets like they exploed or something. So no new traits are required.

Dzierzan:
Yeah, using `ThrowsParticle` generally works, just the facing will be sometimes wrong. It's something to fix in the future IMO, not a big issue right now.

OcelotWalrus (commented):
I think it'd need some playtesting before adding it.

Dzierzan:
Balance will come in another PR, keep it in mind.

OcelotWalrus (approved):

PR #1003, opened 2024-06-23
Map update: Clovergrief (Fixed tiling errors)

SiegeSpud:
Changelog:
-Fixed tiling errors
-Removed unused lightsource from rules.yaml


mods/hv/maps/clovergrief/rules.yaml
```diff
@@ -1,5 +1,5 @@
 World:
-	GlobalLightingPaletteEffect:
+	TintPostProcessEffect:
```
Mailaender, line 2: Can you please keep `GlobalLightingPaletteEffect`. They both do the same but I would like to keep palette effects over modern shaders. https://github.com/OpenHV/OpenHV/commit/c041ad747472294bcc6a816adbcc18ee0e28064a

PR #1007, opened 2024-06-26
Updated Mount Brosnan

OcelotWalrus:
Overhauled Mount Brosnan map:
* made grass less flat with some depth
* added cranes to the main island and some tile variation
* fixed shadows
* updated some lightings


GraionDilach:
You broke the mining locations.

SiegeSpud:
The main problem with this map is its size. The second one is the central island - it should be bigger and the bridges wider. Also the volcano could use some cliffs to justyfy the map name :)

Mailaender:
Some backgrounds. Layout is inspired by https://resource.openra.net/maps/13784/ It was the first multiplayer map. My first experiment with lights. I used it extensively to test the game with bot matches as they do a greedy build with works perfectly here with all the resources. More debug than gameplay value.

Dzierzan:
It basically has a lot of historical value, despite being quite obsolete now.

GraionDilach:
I think this is quite much a Westwoody map. I understand it wasn't the best but I think that despite the layout not being perfect, it's still fine for casual stuff.

OcelotWalrus:
I'm reworking on it but, maybe we could leave the original map for its historical value, and the remake on its side.

Mailaender:
You still need to fix the broken mining spots https://github.com/OpenHV/OpenHV/actions/runs/9678958588/job/26703853937?pr=1007#step:8:130

OcelotWalrus:
I know, I still haven't finished upgrading the map.

PR #1014, opened 2024-06-28
Update Howitzer's power icon

OcelotWalrus:
image


mods/hv/bits/howitzer-icon.yaml
```diff
@@ -0,0 +1,5 @@
+FrameSize: 61,60
```
Dzierzan, line 1: Why such size? Icons use 63x47 image size. Please correct that.

mods/hv/bits/howitzer-icon.yaml
```diff
@@ -0,0 +1,5 @@
...
+FrameAmount: 1
+Offset: 0,0
+Author: Cromha
```
Dzierzan, line 4: This is not made by you. Add Daniel Cook as well.

Dzierzan (changes requested):

Dzierzan (approved):

PR #1015, opened 2024-06-28
Added new Yuruki announcer voice.

Dzierzan:
This PR adds new announcer voice for Yuruki faction. The voices were AI generated using this product: https://www.replicastudios.com/products/text-to-speech

I think they're right now too loud, if you also feel like it, I'll volume down the voices.

Synapol voice (my old rusty voice) will be replaced too in a near future.


Dzierzan:
Apparently we don't get the (C) of the generated voices, they have their own proprietary license terms https://www.replicastudios.com/terms-of-service

So this pull request can be closed now.

PR #1030, opened 2024-07-07
Re-recorded Stealth Tank voices.

Dzierzan:
Moreover:
- Voice line "They won't see me" changed to "Unseen"
- Voice line "Consider it dead" changed to "It's already dead".


OcelotWalrus (approved):
I prefer "Consider it dead" over "It's already dead". Plus, I'd say more "Already dead than It's already dead". Other than that, all sounds good!

PR #1046, opened 2024-07-13
Sound improvement - first batch.

Dzierzan:
- Converted all so far 44 kHz quality sounds to Ogg format for optimalization.
- Recreated or upscaled sounds in CC0 folder.
- Likewise for CC4.0 folder.
- Moved `Freqman_fx376` sound files to CC4.0 as apparently the author changed its license.
- Fixed `fire_effect_X_GregorQuendel` sounds being not audible because wav (now ogg) extension was missing.

You might ask "What does `upscaled` mean for sounds?". Suffice to say, I exported a specific sound as 44 kHz and then I used Wavosaur tool to perform high precision FFT which creates semi high pitch tones. Then High-Pass filter was used to cut off lower tones. Such sound is then merged with the "original" 22 kHz one. I also added some tremble and bass where I felt like it was needed. Basically this method is not perfect, but on average it gives good enough results.


OcelotWalrus (approved):
I didn't check everything, but this sounds good to me.

PR #1055, opened 2024-07-18
Overhaul to Ice River mission

OcelotWalrus:


Dzierzan:
So... patrolling wasn't possible in the end? I am not sure if removing almost all enemy actors was a good idea.

mods/hv/maps/ice-river/map.yaml
```diff
@@ -211,6 +153,37 @@ Actors:
...
+	Actor36: electricpod
+		Owner: Yuruki Industries
+		SubCell: 1
+		Facing: 384
+		Location: 50,52
+	Actor37: saucer
+		Owner: Yuruki Industries
+		Facing: 384
+		Location: 42,55
+	Actor38: flyingmonster
```
Dzierzan, line 183: Nah, don't add force adding critter.

Dzierzan (changes requested):

mods/hv/maps/ice-river/map.yaml
OcelotWalrus: ?

OcelotWalrus:
> So... patrolling wasn't possible in the end? I am not sure if removing almost all enemy actors was a good idea.

It's possible, but anyway the enemy will periodically produce units. But there can be patrolling units if you want.

mods/hv/maps/ice-river/map.yaml
Dzierzan: Remove that critter please.

Dzierzan:
> > So... patrolling wasn't possible in the end? I am not sure if removing almost all enemy actors was a good idea.
> 
> It's possible, but anyway the enemy will periodically produce units. But there can be patrolling units if you want.

I did like some defenders in the first iteration of this mission. I know AI will use them as soon as it activates. You could technically add 2nd enemy player with the same color with brain dead AI (just enemy with no AI logic applied, kinda like Creep faction), so 1st enemy will never use them. A bit hacky, but it could work.

OcelotWalrus:
>  I know AI will use them as soon as it activates

The AI won't, as they're ordered to patrol through the script. I added some patrolling units on this force push.

mods/hv/maps/ice-river/map.yaml
OcelotWalrus: Done

mods/hv/maps/ice-river/map.yaml
```diff
@@ -211,6 +153,42 @@ Actors:
...
+		Owner: Yuruki Industries
+		SubCell: 1
+		Location: 44,49
+		Facing: 493
+	Actor36: electricpod
+		Owner: Yuruki Industries
+		SubCell: 1
+		Facing: 384
+		Location: 50,52
+	Actor37: saucer
```
Dzierzan, line 176: Please, remove the saucer.

Dzierzan (changes requested):

Dzierzan:
image
Two types of MG pods are available.

Dzierzan:
Remove AA tank from Factory and Trade Platform. The enemy faction has only one balloon and you can destroy that with electric pods anyway.

I think Repair Tank could be actually readded.

OcelotWalrus:
Repair Tank hasn't been disabled.

Dzierzan:
Right, I forgot it's in Utility tab.

OcelotWalrus:
Is it good for you @Dzierzan ?

Dzierzan:
AI can build more miners than needed:
image

I noticed patrolling enemy units do not attack back.

OcelotWalrus:
> AI can build more miners than needed:

That's because the periodic productions the AI is ordered to do by the script contain a miner. I don't really see a problem in that in my opinion.

> I noticed patrolling enemy units do not attack back.

That's weird because they do for me. Are you sure you updated your branch?

Dzierzan:
Then the script needs to take in account having already existing miner unit on the battlefield, can you try to update it? 

Yes, I am sure my branch was up to date.

OcelotWalrus:
> Then the script needs to take in account having already existing miner unit on the battlefield, can you try to update it?

I'll make so if a miner already exists (because it hasn't been sent as there's no more avaible resource spots), the enemy won't produce it.

> Yes, I am sure my branch was up to date.

They do attack back for me.

https://github.com/user-attachments/assets/490b3dd7-8019-42a2-a21c-9e79e28462d3

OcelotWalrus:
I understood why they didn't attack you back. When patrolling, units wait a defined amount of time at the waypoint they're in. During that time they can't attack (idk why). So I set that amount to 0 to fix that.

PR #1059, opened 2024-07-20
Added Torpedo Boat unit.

Dzierzan:
- It replaces Medium Boat.
- Medium Boat is kept for potential future use but the unit itself is disabled.

torpedoboat


mods/hv/languages/rules/en.ftl
```diff
@@ -589,6 +589,13 @@ actor-mediumboat =
       Strong vs Water
       Can't attack Pods, Buildings, Vehicles or Air
 
+actor-torpedoboat =
+   .name = Torpedo Boat
+   .generic-name = Boat
+   .description = A boat with torpedo launchers.
+      Strong vs Water
+      Can't attack Pods, Buildings, Vehicles or Air
```
OcelotWalrus, line 597:
```suggestion
      Can't attack Ground and Air
```

mods/hv/languages/rules/en.ftl
```diff
@@ -589,6 +589,13 @@ actor-mediumboat =
       Strong vs Water
       Can't attack Pods, Buildings, Vehicles or Air
 
+actor-torpedoboat =
+   .name = Torpedo Boat
+   .generic-name = Boat
+   .description = A boat with torpedo launchers.
```
OcelotWalrus, line 595:
```suggestion
   .description = A boat with torpedo pods.
```

OcelotWalrus (changes requested):

mods/hv/languages/rules/en.ftl
WilliamWeberBerrutti: I think torpedo launchers is fine. Using torpedo pods can be confusing because there's an unit category called "Pods" already, and the player might think: "Is it launching units I can control?".

mods/hv/languages/rules/en.ftl
WilliamWeberBerrutti: This could be a good change, because Harbors are buildings too.

mods/hv/languages/rules/en.ftl
Dzierzan: Yeah, launcher sounds more correct here.

OcelotWalrus (approved):

PR #1061, opened 2024-07-21
Rebranded Medium Boat as Mercenery Boat

Dzierzan:
- Replaced `Medium Boat` with `Light Boat` in mission `Scientist Rescue`
- Merc Boat is basically a tank on water, so it can shoot targets with ground and vehicle targetable.
- Updated some lua replacing Medium Boat with Torpedo Boat.
- Merc Boat is not buildable (like Merc Tanks). Possible Single Player usage.
- Removed `MediumBoatCannon` weapon.
- Updated AI with Torpedo Boats.


OcelotWalrus:
We discussed with @WilliamWeberBerrutti  and Rampoina on discord, and came up with this idea:

```
I'd propose to change the water units types:
                                                     YI                                           SC
light (equivalent: bike) - Light Boat            - Patrol Boat
heavy (equivalent: mbt) - Medium Boat            - NEW UNIT
torpedo (A-water only) - Dizerzan's torpedo boat - Submarine
```

We need to discuss about that further before changing stuff I think.

Dzierzan:
Honestly I hate adding new units for MP and Mothership/Battlefield was already extended. Hence I prefer rebranding Medium Boat to Merc Boat.

AA aircraft will be most likely the last ones I am willing to add because we're again facing the same issue, too many units.

OcelotWalrus:
> too many units.

I don't see that as an issue. But I understand.

mods/hv/rules/ships.yaml
```diff
@@ -122,15 +122,11 @@ MEDIUMBOAT:
...
-	DetectCloaked:
-		Range: 5c0
-		DetectionTypes: Underwater
-	RenderDetectionCircle:
 	Turreted:
 		TurnSpeed: 60
 		Offset: 150,0,100
 	Armament:
-		Weapon: MediumBoatCannon
+		Weapon: ^Cannon
```
Mailaender, line 129: That is not a real weapon but a template used for inheritance.

OcelotWalrus: It is actually a real. The merctank uses it, as the mbt (mbt uses an other, but it's exactly the same, with just a different bullet image)

Mailaender: Fixed the mercenary tank in https://github.com/OpenHV/OpenHV/commit/077fe5608fc08d3d68ee6eb55e9ea5c15aaf19f5

OcelotWalrus: Oh my bad then.

OcelotWalrus:
> > too many units.
> 
> I don't see that as an issue, it just brings more gameplay options.

I just wanna add that this reorganization of water units would just add 1 new unit.

PR #1064, opened 2024-07-21
Enlarged Lightning Tank

Dzierzan:
image


OcelotWalrus (approved):

Mailaender:
They do overlap as they are larger than cell size, but it is actually worse for the APC.
grafik

OcelotWalrus:
Maybe they would need a footprint? Just and idea like that.

Mailaender:
Only buildings have footprints.

OcelotWalrus:
But how other mods do? For example, D2K have units bigger than a cell, and they don't overlap.

GraionDilach:
They do overlap. Other mods also ignore overlapping.

The engine lacks support for multicell units anyway.

PR #1072, opened 2024-07-25
Builder SC has different art now.

Dzierzan:
Commit da005d2d7dcaf63bcbaf486b6085ad887938e4ff made both builders looking the same despite there were at least two more differently looking builders to use.


OcelotWalrus:
Just to tell you about it, the builder5 spritesheet has an error:

builder5

The last sprite has green and gray colors swapped.

Dzierzan:
Thanks for noticing it! It's been there since the beginning of this project :D.

PR #1073, opened 2024-07-25
Added force field effect artwork.

Dzierzan:
However `GrantExternalConditionPower ` doesn't support image effect. So for time being, only art is submitted.


OcelotWalrus (approved):
Ain't implemented yet, but looks nice! I like it.

Dzierzan:
GIF for reference:
forceshieldeffect

OcelotWalrus:
All the effects you're adding (comlink, force field etc...) are really nice. It really enhances the game's visuals.

GraionDilach:
Looks too isometric and not enough topdown IMO.

OcelotWalrus:
> Looks too isometric and not enough topdown IMO.

Maybe. But we need to see how it actually looks in-game to be sure.

Mailaender:
I can't easily extend `GrantExternalConditionPower` for a sprite effect. However [GrantExternalConditionPower](https://openhv.readthedocs.io/en/latest/release/traits/#grantexternalconditionpower) already has support for an `active` sequence on the targeted actors.

GraionDilach:
You missed the discussion from chat but I'll resummarize.

The `active` sequence is actually the same functionality what upstream's `WithSupportPowerActivationAnimation` trait does.
`DetonateWeaponPower` with a `GrantExternalConditionWarhead` can provide the needed functionality, but  it would need a restructuring of the setup and shifting footprint range for warhead spread range.

Mailaender:
Ah `active` is on the building and not on the unit.

GraionDilach:
I'm not sure why would you call anything of that as fake, though. I mean, the DWP weapon's explosion having a CreateEffectWarhead along is where this would be used.

PR #1081, opened 2024-07-27
Change both Hacker and Countermeasure tanks to Utility tab

WilliamWeberBerrutti:
- Disables both land and naval minelayers
- Moves both Hacker and Countermeasure tanks to Utility tab

### DIFF
--- a/mods/hv/rules/vehicles.yaml
+++ b/mods/hv/rules/vehicles.yaml
@@ -566,7 +566,7 @@ ECMTANK:
 	Selectable:
 		Class: ecmtank
 	Buildable:
-		Queue: Tank
+		Queue: Utility
 		Prerequisites: ~factory.sc, techcenter, trader
 		BuildPaletteOrder: 60
 		Description: actor-ecmtank.description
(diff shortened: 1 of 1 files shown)


mods/hv/rules/vehicles.yaml
```diff
@@ -831,7 +831,7 @@ HACKERTANK:
 	Inherits: ^TrackedVehicle
 	Inherits@AutoTarget: ^AutoTargetGroundAssaultMove
 	Buildable:
-		Queue: Tank
+		Queue: Utility
 		BuildPaletteOrder: 60
 		Prerequisites: ~factory.yi, radar, trader
```
Mailaender, line 836: This makes them invisible before a factory is built which is irritating.

WilliamWeberBerrutti: Oh didn't know that! How to fix it?

OcelotWalrus: Just remove the `~`.

Mailaender: That will show the internal string `factory.yi` in the build palette.

OcelotWalrus: Oh yeah right. But anyway, I don't see why they're being moved to the utility tabs, they don't belong there.

Dzierzan:
`Disables both land and naval minelayers`

No, I do not accept that. You don't use them in MP? OK, so don't use them, but AI uses them nicely. So I see no problem here.

GraionDilach:
> Disables both land and naval minelayers

This is already dropped from the PR.

WilliamWeberBerrutti:
Yea this experimental PR was a bad idea. Thanks for the feedbacks! Closing it!

PR #1082, opened 2024-07-28
Mentioned Synapol submarines use in its description

OcelotWalrus:


mods/hv/languages/rules/en.ftl
```diff
@@ -49,6 +49,7 @@ faction-synapol =
 
     Faction Variations:
         - Uses helicopters as air units
+        - Uses submarines as advanced naval units
```
Mailaender, line 52: This is unbalanced as there is no counterpart statement on the other faction.

OcelotWalrus: What do you want me to say? There's no counterpart, I think.

OcelotWalrus: If not saying "Uses boats as advanced naval units" in Yuruki's description.

Mailaender: Yeah that is lame.

PR #1083, opened 2024-07-28
Fix units hiding behind high walls

### LINKED ISSUE #1078: High walls block projectiles
Units behind top side of walls can't be directly damaged by missiles (potentially other attacks too) and units on that position have their projectiles blocked.

Screenshot for illustration:

Screenshot_20240726_173254

OcelotWalrus:
Fixes #1078.

### DIFF
--- a/mods/hv/rules/buildings.yaml
+++ b/mods/hv/rules/buildings.yaml
@@ -2039,7 +2039,7 @@ WALLT:
 	Inherits@Shape: ^1x1Shape
 	Building:
 		Dimensions: 1,1
-		Footprint: =
+		Footprint: x
 	Interactable:
 		Bounds: 1228, 1228
 	WithSpriteBody:
@@ -2050,7 +2050,7 @@ WALLTL:
 	Inherits@Shape: ^1x1Shape
 	Building:
 		Dimensions: 1,1
-		Footprint: =
+		Footprint: x
 	Interactable:
 		Bounds: 1228, 1228
 	WithSpriteBody:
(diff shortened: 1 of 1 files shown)


Dzierzan:
I do not approve this change. I'd rather find more proper solution than making passable terrain unpassable.

OcelotWalrus:
> I do not approve this change. I'd rather find more proper solution than making passable terrain unpassable.

image

Dzierzan:
Check the harbor map, you'll understand why this idea is not that great. Try to do something with hitbox instead.

OcelotWalrus:
Sorry but I don't see why.

Dzierzan:
South and north passages are 3 tile wide. If you that PR is merged, then west and east entries will be only 2 tiles wide.

image

OcelotWalrus:
Oh yeah I see. I'll try to do something with the hitbox.

OcelotWalrus:
@Mailaender the problem is that whether it has no hitbox, or a small one, it'll never look good with bullets firing on/in it. I don't think units should be able to go on these tiles. Maps like harbour that Dzierzan mentioned would then need to be adapted to that change.

OcelotWalrus:
Actually @Mailaender, removing the hitbox doesn't cause weird visual bugs with bullets passing through, as these tiles are drawn over them anyway.

Mailaender:
Still broken
grafik

mods/hv/rules/buildings.yaml
```diff
@@ -2036,7 +2036,6 @@ VGATE:
 
 WALLT:
 	Inherits: ^Wall
-	Inherits@Shape: ^1x1Shape
 	Building:
 		Dimensions: 1,1
 		Footprint: =
```
Mailaender, line 2041:
```suggestion
		Footprint: x
```
Just disallow units hiding there.

OcelotWalrus: Okay!

PR #1096, opened 2024-07-31
Balance Battleship

WilliamWeberBerrutti:
**NOTE:** This PR ideally should be merged together with [Balance Mothership](https://github.com/OpenHV/OpenHV/pull/998).

### DIFF
--- a/mods/hv/rules/aircraft.yaml
+++ b/mods/hv/rules/aircraft.yaml
@@ -811,21 +811,22 @@ BATTLESHIP:
 	Buildable:
 		Queue: Aircraft
 		BuildPaletteOrder: 100
-		Prerequisites: ~starport.yi, techcenter, ~disabled
+		BuildLimit: 1
+		Prerequisites: radar, trader, techcenter, ~starport.yi
 		Description: actor-battleship.description
 	UpdatesPlayerStatistics:
 		AddToArmyValue: true
 	Health:
-		HP: 100000
+		HP: 250000
 	Armor:
 		Type: Heavy
 	RevealsShroud:
-		Range: 12c0
-		MinRange: 4c0
+		Range: 16c0
+		MinRange: 8c0
 		Type: GroundPosition
 		RevealGeneratedShroud: False
 	RevealsShroud@Hacked:
-		Range: 6c0
+		Range: 8c0
 		Type: GroundPosition
 	Aircraft:
 		TurnSpeed: 20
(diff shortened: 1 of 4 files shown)


mods/hv/rules/aircraft.yaml
```diff
@@ -811,12 +811,13 @@ BATTLESHIP:
 	Buildable:
 		Queue: Aircraft
 		BuildPaletteOrder: 100
-		Prerequisites: ~starport.yi, techcenter, ~disabled
+		BuildLimit: 1
+		Prerequisites: radar, trader, techcenter
```
OcelotWalrus, line 815: You made it available to all factions.

OcelotWalrus: You must also add the `~starport.x` prerequisite or it'll still appear even if the player has no starport (if I'm right)

mods/hv/rules/aircraft.yaml
```diff
@@ -811,17 +811,18 @@ BATTLESHIP:
...
 	Health:
-		HP: 100000
+		HP: 250000
 	Armor:
 		Type: Heavy
 	RevealsShroud:
-		Range: 12c0
-		MinRange: 4c0
+		Range: 16c0
+		MinRange: 6c0
```
OcelotWalrus, line 825:
```suggestion
		MinRange: 8c0
```

From what I've seen, `MinRange` is always the halved value of `Range`.

OcelotWalrus (changes requested):

mods/hv/weapons/ballistics.yaml
```diff
@@ -351,16 +351,11 @@ BunkerMortarGun:
...
+	ReloadDelay: 110
 	Range: 16c0
 	MinRange: 8c0
-	# IDEA: Maybe balance as a single attack, then think on varying its bursts! It should be easier this way!
-	# # Burst: 2
-	# # BurstDelays: 8
 	Burst: 2
 	BurstDelays: 0
-	# Burst: 8
-	# BurstDelays: 8
```
Mailaender: Please merge this into your previous commit.

mods/hv/rules/aircraft.yaml
WilliamWeberBerrutti: It's not always the case, but I think we should change to always be the half of `Range` for all units to be consistant.

PR #1097, opened 2024-08-01
Differentation between the two factions' scouts

OcelotWalrus:
* the Balloon has now a higher range (18->24)
* the Saucer has now a higher speed (70->80)


WilliamWeberBerrutti:
I'd balance scouts differently:

- Synapol scout have more HP (maybe +5%) and less speed (maybe -5%)
- Yuruki scout have less HP (maybe -5%) but more speed (maybe +5%)

OcelotWalrus:
Yeah but, the way they look (their sprites) it's more like synapol (the balloon) has less HP and more speed.

PR #1105, opened 2024-08-03
Fix Oresmelt playing animation when damaged

### LINKED ISSUE #1104: Ore Smelters from both Factions incorrectly animate when severely damaged

OcelotWalrus:
Closes #1104


WilliamWeberBerrutti:
I don't think you should disable cash generation when damaged.

PR #1108, opened 2024-08-03
Enlarged map Fight At The River

OcelotWalrus:


mods/hv/maps/river-fight/map.yaml
```diff
@@ -35,102 +35,11 @@ Players:
 		Enemies: Creeps
 	PlayerReference@Multi1:
 		Name: Multi1
-		AllowBots: False
```
Mailaender, line 38: Bots still can't use landing craft. You could try to do a custom AI for this map that only attacks by air and sea, though.

OcelotWalrus: It's the map editor that must have removed that.

PR #1070, opened 2024-07-23
Added Palm Tree decoration actor

OcelotWalrus:
_Partially fixes #1041._

image

PR #1109, opened 2024-08-03
Added palm trees.

Dzierzan:
Supersedes #1070

image


Dzierzan:
Leaves were toned down by -25%

OcelotWalrus (approved):
Looks amazing!

I'd be amazed if you could also make palm tree tiles:
image

PR #1114, opened 2024-08-04
Recode ImpactAirWarhead to ValidateTriggerWarhead.

GraionDilach:
ImpactAirWarhead was inherited from AS, but due to bitrot and confusion over the initial approach, the reasons stopped being clear enough on why it got implemented in the first place.

This refactor shifts all trigger validation logic to the abstract base class from the downstream usecases, to promote the trigger validation as the primary function and downplay the Air targettype grant as an implementation detail.

Also dropped ImpactActors tag to explicitly rely on TargetType filtering alone.

TO-DO: Write case study properly explaining the underlying reason.

### DIFF
--- a/OpenRA.Mods.HV/Warheads/FireRadiusWarhead.cs
+++ b/OpenRA.Mods.HV/Warheads/FireRadiusWarhead.cs
@@ -13,13 +13,12 @@ using System;
 using System.Linq;
 using OpenRA.GameRules;
 using OpenRA.Mods.Common.Traits;
-using OpenRA.Mods.Common.Warheads;
 using OpenRA.Traits;
 
 namespace OpenRA.Mods.HV.Warheads
 {
 	[Desc("Fires a defined amount of weapons with their maximum range in a wave pattern.")]
-	public class FireRadiusWarhead : ImpactAirWarhead, IRulesetLoaded<WeaponInfo>
+	public class FireRadiusWarhead : ValidateTriggerWarhead, IRulesetLoaded<WeaponInfo>
 	{
 		[WeaponReference]
 		[FieldLoader.Require]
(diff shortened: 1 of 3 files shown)


GraionDilach:
That FRW pos usage is a bug, I'll need to review what was I thinking there.

GraionDilach:
Might need to split this one into two, because I got carriede away wrt FireRadiusWarhead.

OpenRA.Mods.HV/Warheads/FireRadiusWarhead.cs
```diff
@@ -76,31 +64,24 @@ public override void DoImpact(in Target target, WarheadArgs args)
 
 			for (var i = 0; i < amount; i++)
 			{
-				var radiusTarget = Target.Invalid;
-
 				var rotation = WRot.FromFacing(i * offset);
-				var targetpos = epicenter + new WVec(weapon.Range.Length, 0, 0).Rotate(rotation);
-				var tpos = Target.FromPos(new WPos(targetpos.X, targetpos.Y, map.CenterOfCell(map.CellContaining(targetpos)).Z));
+				var radiusTargetVector = epicenter + new WVec(weapon.Range.Length, 0, 0).Rotate(rotation);
```
Mailaender, line 68: Not a vector but a position.

OpenRA.Mods.HV/Warheads/FireRadiusWarhead.cs
```diff
@@ -76,31 +64,24 @@ public override void DoImpact(in Target target, WarheadArgs args)
...
-					tpos = Target.FromPos(new WPos(targetpos.X, targetpos.Y, map.CenterOfCell(map.CellContaining(targetpos)).Z + altitude.Length));
-				}
+				// allow maintaining Z axis in airburst chain
+				var radiusTargetAltitude = TransferAltitude
+					? map.CenterOfCell(map.CellContaining(radiusTargetVector)).Z + map.DistanceAboveTerrain(targetPosition).Length
+					: map.CenterOfCell(map.CellContaining(radiusTargetVector)).Z;
 
-				if (weapon.IsValidAgainst(tpos, firedBy.World, firedBy))
-					radiusTarget = tpos;
+				var radiusTargetPosition = Target.FromPos(new WPos(radiusTargetVector.X, radiusTargetVector.Y, radiusTargetAltitude));
```
Mailaender, line 75:
```suggestion
				var radiusTarget = Target.FromPos(new WPos(radiusTargetVector.X, radiusTargetVector.Y, radiusTargetAltitude));
```

Issue #1115, opened 2024-08-04
Add an `ActivationDelay` field to both DropPodsPower and TeleportPower


OcelotWalrus:

OcelotWalrus:
If it's not too much asking Mailaender, would you mind adding that functionality to the engine this weekend or next one? One of my PR misses that functionality to be merged. I don't think it's too complicated to implement that if I'm right.

Mailaender:
That would feel awkward. Support powers need to be instant.

PR #1116, opened 2024-08-04
Enhanced the Support Powers behaviors

### LINKED ISSUE #1112: Add a warning time/delay for the support powers.
It would be nice if the support powers would give the player a warning before they hit so they have a bit of time to react:

- **Airstrike**: like other games the direction the planes are coming from could be shown to the player with lines.
- **Droppods**: the area the droppods are going to drop could be shown a few seconds before they do with a circle or something like that (as long as you have vision of that area)
- **Howitzer**: some kind of target could be shown where the howitzer is pointing at. (as long as you have vision)
- **Televator power**: the delay where the units are appearing but aren't targettable could be increased. (one really strong strategy is to teleport bomb pods/flamers directly on top of the base)

OcelotWalrus:
Fixes #1112, ~requires #1115 to be fixed.~

* Make support powers that ain't instant (airstrike, nuke, railgun) place a beacon with an arrow in the player's color, at the target location of that support power (only displays to teammates).
* Gave the railgun an activation delay of 10 ticks
* ~Gave the drop pods power an activation delay of 10 ticks  # waiting for requirements~
* ~Gave the televator power an activation delay of 10 ticks  # waiting for requirements~


GraionDilach:
> Also requires the AttackOrderPower to support the DisplayBeacon trait.

That's won't happen in the likely future. The main reason why this is missing from the engine is because `AttackOrderPower` cannot estimate how fast the projectile is, due to the way projectiles work (arcing vs missiles vs instanthit) and/or of it's using multistages (via using any of the `Fire*WeaponWarheads`).

OcelotWalrus:
Yeah I see. Do you know any other solution to indicate a location?

mods/hv/rules/buildings.yaml
```diff
@@ -1587,7 +1597,6 @@ UPLINK:
 		SelectTargetSpeechNotification: SelectTarget
 		SelectTargetTextNotification: notification-select-target
 		IncomingSpeechNotification: OrbitalRailgunFired
-		ActivationDelay: 0
```
GraionDilach, line 1590: Do we really need this additional delay?

I mean, the projectile speed and the fact that it falls from a great height already gives it a more than enough delay as-is.

It might be better if there would be a dummy beacom effect after activation to instantly show where it will trigger instead of this change.

OcelotWalrus: Well there're two problems: adding a beacon to the uplink's support power makes the game crash on activation, and as soon as the weapon is activated, the beacon disappears: so if we give it no delay, the beacon never shows.

GraionDilach: That's a lengthy way of saying "I don't know how to fake a beacon".

The crash needs a log and I'll might wrap the dummy beacon effect during the weekend.

Mailaender:
I don't think we need this as there is already a radar triangle showing the location.

OcelotWalrus:
First, the triangle can mean anything, and secondly, it's nice to have a precise strike target shown for teammates to follow on the strike.

PR #1137, opened 2024-08-08
Update Howitzer's icon credit attributions

OcelotWalrus:
From now on, I'm using my name for credit attributions purposes.


mods/hv/bits/howitzer-icon.yaml
```diff
@@ -1,5 +1,5 @@
 FrameSize: 63,47
 FrameAmount: 1
 Offset: 0,0
-Author: Daniel Cook, Cromha
+Author: Daniel Cook, Jimmy L. Miles (Cromha)
```
Mailaender, line 4: Put your real name (nickname) in credits and go with your real name only here. We also don't credit Daniel Cook with (danktheduck).

PR #1142, opened 2024-08-09
Don't play animation on orepurifier/smelter when it's being sold

### LINKED ISSUE #1125: A Frame of Ore Smelters appear briefly after selling them

OcelotWalrus:
Fixes #1125.

### DIFF
--- a/mods/hv/rules/buildings.yaml
+++ b/mods/hv/rules/buildings.yaml
@@ -1063,7 +1063,7 @@ ORESMELT:
 		RequiresCondition: !build-incomplete && !disabled
 	WithIdleAnimation:
 		Interval: 40
-		RequiresCondition: !build-incomplete && !disabled && !burning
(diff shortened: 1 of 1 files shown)


OcelotWalrus:
Should I move the `sold` condition to the oresmelter/purifer only? So the workflows are happy :D

mods/hv/rules/defaults.yaml
```diff
@@ -602,6 +602,8 @@
 	ActorLostNotification:
 		Notification: BuildingLost
 		TextNotification: notification-building-lost
+	GrantConditionOnSell:
+		Condition: sold
```
Mailaender, line 606: This is duplicating https://github.com/OpenHV/OpenHV/blob/42a4b0f82cc504806c3b8724d22c0a5f44b374a8/mods/hv/rules/defaults.yaml#L599-L600

PR #1136, opened 2024-08-08
Added a new prop

OcelotWalrus:

PR #1145, opened 2024-08-11
Added prop18.

Dzierzan:
Supersedes #1136

prop18


OcelotWalrus (commented):
The drilling animations look terrible in my honest opinion (it drills down way too fast). You should add outlines to the basements. The shadow should also be drawn over the center basement, logically. The shadow also looks odd but maybe that's just me.

Dzierzan:
To be honest, it doesn't look like a drill to me, but more like a hammer-thingy. Hence pounding the hole, for whatever reason you might image. So your guess is as good as mine, but drill... I don't see it.

I prefer foundation without outline.

The shadow is correct.
image

OcelotWalrus:
At least don't make the animation when it goes down that harsh, it doesn't feel natural.

OcelotWalrus:
I'd just add one more frame though, at the middle of the descent.

Dzierzan:
I can add that one frame, but I'd avoid using particles.

Dzierzan:
Alright, I actually re-added your original idea of `active` sequence, as I forgot pre-placed props would have synchronized animation, which would look odd.

I also changed a bit animation. See if you like it now.

Dzierzan:
Ehh... a small desync with shadow, gotta fix it. Barely noticable, but I see it xD

Dzierzan:
prop18-2

OcelotWalrus (approved):

PR #1148, opened 2024-08-13
Added prop19.

Dzierzan:
prop19


mods/hv/rules/props.yaml
```diff
@@ -378,6 +378,53 @@ PROP18:
...
+	Health:
+		HP: 60000
+	HitShape:
+		UseTargetableCellsOffsets: False
+		TargetableOffsets: -100,-412,0, -100,412,0, -800,-412,0, -800,412,0
+		Type: Rectangle
+			TopLeft: -1024, -424
+			BottomRight: 1024, 1524
+	WithSpriteBody@Animation:
+		Sequence: null
```
Mailaender, line 399: Can you explain why you did this?

Dzierzan: Yeah, I assumed that if I didn't do it, idle animation could potentially cover other animations.

OcelotWalrus: I think it's what he uses to make the animations random.

GraionDilach: We already had issues with having multiple WSBs (#148 triggering https://github.com/OpenRA/OpenRA/pull/18574#issuecomment-683272807 leading to #177), so this is a bad setup.

Use z offsets.

Dzierzan: So you want to render idle 3 times?

GraionDilach: It would be better if you'd use WithIdleOverlay+GPC from #1150 for the lights and the animation, instead of the WSB+WIA combination.

Mailaender: Animations tight in with conditions sounds like too much overhead. I'd rather write a new animation trait.

GraionDilach: I don't see how would that new animation trait look like. Sure, you can just copy-paste WIO and add some sort of a play-once-every-random-amount-of-frames support, but I'm not convinced that such code duplication is the correct path.

PR #1158, opened 2024-08-21
Remove Broker Pod deployable ability and prevent bots from building them

WilliamWeberBerrutti:


GraionDilach:
Should aqlso delete the sequence definition at least to not waste pixels on the generated internal sprite sheets.

OcelotWalrus (commented):
Why preventing the bots from building broker pods?

WilliamWeberBerrutti:
> Why preventing the bots from building broker pods?

Because, with the ability removal, they don't know to use them to steal money, as far as I noticed.

GraionDilach:
It's likely the same issue as upstream's takeaway from medics/mechanics then - the AI spamming attack-moves and relying on autotargeting thenafter is too crude to be effectively combined with such a specific attacking usecase.

PR #1159, opened 2024-08-21
Added range circles to Nuke and Railgun support powers

OcelotWalrus:
Added a range circle to both Nuke and Railgun support powers, to show the damage range to the player.

### DIFF
--- a/mods/hv/rules/buildings.yaml
+++ b/mods/hv/rules/buildings.yaml
@@ -1506,6 +1506,8 @@ SILO:
 		SelectTargetTextNotification: notification-select-target
 		IncomingSpeechNotification: NukeIncoming
 		OrderName: Nukem
+		CircleRanges: 9c0
+		CircleColor: FFFFFF
 	SupportPowerChargeBar:
 	Power:
 		Amount: -150
@@ -1588,6 +1590,7 @@ UPLINK:
 		SelectTargetTextNotification: notification-select-target
 		IncomingSpeechNotification: OrbitalRailgunFired
 		AirburstAltitude: 40c0
+		TargetCircleRange: 3c0
 		Weapon: RailgunSetup
 
 STORAGE:


mods/hv/rules/buildings.yaml
```diff
@@ -1588,6 +1590,7 @@ UPLINK:
 		SelectTargetTextNotification: notification-select-target
 		IncomingSpeechNotification: OrbitalRailgunFired
 		AirburstAltitude: 40c0
+		TargetCircleRange: 5c0
```
Mailaender, line 1593:
```suggestion
		TargetCircleRange: 3c0
```

mods/hv/rules/buildings.yaml
```diff
@@ -1506,6 +1506,8 @@ SILO:
 		SelectTargetTextNotification: notification-select-target
 		IncomingSpeechNotification: NukeIncoming
 		OrderName: Nukem
+		CircleRanges: 25c0
```
Mailaender, line 1509: This is more than the entire screen.
grafik

```suggestion
		CircleRanges: 6c0
```
maybe

mods/hv/rules/buildings.yaml
OcelotWalrus: I actually tested by comparing the range circle and the red damage circle with the combat geometry overlay and it fits correctly.

mods/hv/rules/buildings.yaml
OcelotWalrus: I know it's huge but as with the uplink, it fits the damage zone. It even is a bit smaller.

OcelotWalrus: On my screen it doesn't.

Mailaender: You are including the fall off.

PR #1171, opened 2024-08-27
Increase Combat Units Cost and Tweak AI

WilliamWeberBerrutti:
Motivations:

- Aids on performance issues when there's a lot of units in the match
- By reducing the combat units amount, it's easier to micro, being a little more important for the game
- Tweak AI with the new cost changes and some other changes like building at least 2 Storages so it's harder to kill it

The cost increase proportion is +100%. Units not affected are: Miner, Builders, Recon Tank, Money Transport

### DIFF
--- a/mods/hv/rules/bots.yaml
+++ b/mods/hv/rules/bots.yaml
@@ -16,9 +16,11 @@ Player:
 	BaseBuilderBotModule@RogueAI:
 		RequiresCondition: enable-rogue-ai
 		MinimumExcessPower: 40
-		MaximumExcessPower: 200
-		AdditionalMinimumRefineryCount: 0
-		NewProductionCashThreshold: 10000
+		MaximumExcessPower: 500
+		InititalMinimumRefineryCount: 1
+		AdditionalMinimumRefineryCount: 1
+		PlaceDefenseTowardsEnemyChance: 0
+		NewProductionCashThreshold: 7000
 		WaterTerrainTypes: Littoral
 		ConstructionYardTypes: base, base2
 		PowerTypes: generator
@@ -30,7 +32,7 @@ Player:
 		DefenseTypes: turret, turret2, aaturret, aaturret2, bunker, bunker2
 		BuildingLimits:
 			base: 1
-			storage: 1
+			storage: 3
 			module: 4
 			module2: 4
 			factory: 4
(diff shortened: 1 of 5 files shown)


OcelotWalrus (approved):
The code's good and the idea too. @Rampoina Cyantusk and me playtetsed it.

mods/hv/rules/bots.yaml
```diff
@@ -64,39 +72,47 @@ Player:
...
+			field: 12000
+			silo: 13000
+			uplink: 13000
 	BuildingRepairBotModule:
 		RequiresCondition: enable-rogue-ai
 	SquadManagerBotModule@RogueAI:
 		RequiresCondition: enable-rogue-ai
 		SquadSize: 20
-		ExcludeFromSquadsTypes: miner, builder, builder2, tanker1, tanker2, minelayer, ecmtank, technician, broker, mineship, mineship2, blaster
+		ExcludeFromSquadsTypes: miner, builder, builder2, tanker1, tanker2, minelayer, technician, broker, mineship, mineship2, blaster, flamer
```
Mailaender, line 107: The flamer is a regular attack unit not handled by other bot modules.

WilliamWeberBerrutti: So the idea is to remove flamer there? So the Flamer Pod is going to be included on squads? It shouldn't, in my opinion.

Mailaender: Then write a bot module for them, or they will stay in base.

mods/hv/rules/bots.yaml
```diff
@@ -16,9 +16,15 @@ Player:
...
-		AdditionalMinimumRefineryCount: 0
-		NewProductionCashThreshold: 10000
+		MaximumExcessPower: 500
+		InititalMinimumRefineryCount: 1
+		AdditionalMinimumRefineryCount: 1
+		PlaceDefenseTowardsEnemyChance: 0
+		StructureProductionInactiveDelay: 0
+		StructureProductionActiveDelay: 5
+		StructureProductionRandomBonusDelay: 0
+		StructureProductionResumeDelay: 5
```
Mailaender, line 26: These are typically added for performance reasons so not all bots compute at the same time. Please don't set it to 0 or very low values.

WilliamWeberBerrutti: What values do you suggest?

Mailaender: Just remove them. They are arbitrary.

PR #1181, opened 2024-08-31
Added new tileset *beautiful*: Black Sand

OcelotWalrus:
I got the idea from Iceland's actual black sand.


Dzierzan:
I guess this works. It's a simple reskin, but since it's a reskin of colors, I've got nothing against the art itself.

Looks like burn ground to me.

Mailaender:
You can still give it a different texture.

PR #1183, opened 2024-09-01
Fixed some incorrectly placed high wall pieces in Swampy Frontier map.

Dzierzan:
And adjusted facings for critter.


Mailaender:
It is really hard to spot the difference. My preference would be to remove the walls in the middle entirely.

PR #1184, opened 2024-09-02
Change Planes and Speeders movement pattern and basic air weaponry

WilliamWeberBerrutti:
Motivation:

- Makes Yuruki's planes more useful in late game masses (since they expose themselves too much while engaging and it's very hard to micro them properly, specially with lag)
- Unifies both Helicopter and Plane's anti-air and anti-ground weapons
- Rebalances slightly Speeders and Planes (so they have roughly the same DPS as Banshees and Helicopters like before)


WilliamWeberBerrutti:
I'll close this PR since I intend to do more changes with air.

PR #1185, opened 2024-09-04
Make Force Field affect buildings

WilliamWeberBerrutti:
Motivation:

- Gives an option for Yuruki players to counter Howitzer attacks, specially if it's aimed at critical buildings like Mining Towers
- Gives an option to protect against snipes of important buildings like Storages and Main Bases


OcelotWalrus (commented):
Interesting! I like that idea.

PR #1196, opened 2024-09-11
Added AA aircraft units.

Dzierzan:
There's no balance here yet. Up to you @Rampoina @WilliamWeberBerrutti  to balance it.


WilliamWeberBerrutti:
We'll balance on another PR, probably on the asymmetry one. For me it's good to go!

PR #1202, opened 2024-09-13
Added Observer unit.

Dzierzan:
- Observer replaces Saucer.
- Observer is a copter unit, made for Synapol.
- Moved Balloon unit to Yuruki.
- Disabled Saucer unit.

observer2


Mailaender:
Please also update https://github.com/OpenHV/OpenHV/blob/c568025a16acfb8390de89fcb9afc10a91191cae/mods/hv/rules/world.yaml#L499-L510

PR #1214, opened 2024-09-24
Fix ECM Tank targetting other ECM Tanks

### LINKED ISSUE #1213: Enemy ECM Tanks are not targetted by own ECM Tank
Screenshot_20240924_162541

I'd expect ECM Tanks to target enemy ECM Tanks.

WilliamWeberBerrutti:
Fix #1213.


Mailaender:
That is on purpose.

WilliamWeberBerrutti:
> That is on purpose.

Because of some engine peculiarity?

Mailaender:
Both hacker and ECM tank are immune to their own weaponry for consistency.

PR #1219, opened 2024-10-24
Added a Marine AI to Naval Conquest maps

OcelotWalrus:
I found some time to wrap up a "Marine AI", so that players can play offline on Naval Conquest maps (Split Three & Fight At The River).

I suggest you to spectate a game between bots to see how they handle the play.
Feel free to suggest any kind of crap to enhance it!

### DIFF
--- a/mods/hv/maps/river-fight/map.yaml
+++ b/mods/hv/maps/river-fight/map.yaml
@@ -29,13 +29,11 @@ Players:
 		Enemies: Multi0, Multi1
 	PlayerReference@Multi0:
 		Name: Multi0
-		AllowBots: False
 		Playable: True
 		Faction: Random
 		Enemies: Creeps
 	PlayerReference@Multi1:
 		Name: Multi1
-		AllowBots: False
 		Playable: True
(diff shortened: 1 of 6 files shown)


mods/hv/maps/river-fight/rules.yaml
```diff
@@ -12,6 +12,316 @@
 		Palette: cloud
 		Speed: 20, 25
 
+Player:
+	GrantConditionOnBotOwner@WaterCampaignBot:
+		Condition: enable-watercampaign-ai
+		Bots: watercampaign
```
Mailaender, line 18: campaign?

mods/hv/maps/river-fight/map.yaml
```diff
@@ -14,7 +14,7 @@ Bounds: 1,1,148,130
 
 Visibility: Lobby
 
-Categories: Conquest
+Categories: Naval Conquest
```
Mailaender, line 17: This is the game type not the map type. Think of it as Capture the Flag vs Deathmatch.

PR #1225, opened 2024-10-31
Switched Yuruki and Synapol factions

SeanScherer:
.. , as in my mind, this better matches their background, Logo and names. I'm assuming here that a least most new players would feel this way as well.

Switched all short-name (sc, yi) references in /rules, and all "Availability:" references in the Encyclopedia references.

Missions / campaign is left untouched as of now (but I was looking at delving in there anyhow).

### DIFF
--- a/mods/hv/languages/rules/en.ftl
+++ b/mods/hv/languages/rules/en.ftl
@@ -24,31 +24,6 @@ faction-yuruki =
     an aggressive expansion throughout the solar system.
 
     Faction Variations:
-        - Uses fighter jets as air units
-
-    Special Units:
-        - Shocker Pod
-        - Blaster Pod
-        - Sniper Pod
-        - Gatling Bike
-        - Hacker Tank
-        - Lightning Tank
-        - Stealth Tank
-        - Battleship
-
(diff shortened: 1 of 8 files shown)


SeanScherer:
A couple sentences explanation, since it might not / probably isn't clear from above: 
My intention is to match the aesthetics - that is the visuals / sprites, as well as weapons better to the "vibe" of the two factions.

So, to expand on that just a little, just the name "Synapol" to me implies "subversion", subtle, sophisticated; and their background as a interplanetary mega-corp production company suggests high-tech to me. (So that fits eg with the more rounded style Module, and the elektro/lightning weapons).

Yuruki Industries on the other hand (coming from Colonial expantion) sounds really grounded and - thinking of Colonial marines, more gung-ho and present day tech. So that would imply more conventional weaponry, ie the rockets, and more "basic" architecture (eg the Module).

( I think there's still room for shifting some stuff this way and that, but just this basic "label switch" leaves me far more satisfied and less jarred compared to what I'm expecting from the factions :P . )

Mailaender:
The factions are already established in previous releases.

PR #1235, opened 2024-12-23
Fix Ore on Black Sand, and Miner being deployable on Red Snow

### LINKED ISSUE #1234: Ore can't be used on all expected terrain
There are some terrain types that will not allow placement of Ore or deployment of Miner, even though it would be expected to work:

Tech, Black Sand, Stone, Ice and possibly Plant.

For Tech (and Stone and Ice if simply added to the valid Ore terrain types in world.yaml) placement of Ore in the MapEditor does not work.

For Black Sand it appear to be placeable, but then is not displayed in game.

Red Snow shows Ore, but Miner will not deploy.

Plant shows Ore, but Miner is not able to drive over/to Plant Terrain.

These issues obviously have different causes, but I've collected them since they work out having the same effect (Ore not usable on a certain Terrain type).

Miners-Terrain

SeanScherer:
Change world.yaml and buildings.yaml to add entries for "Black Sand" (as a valid Ore Terrain type) and "Red Snow" (as a buildable terrain type for the Miner) - thus fixing ore not being usable on these two terrain types.

Placing Ore on "Tech" terrain will still not work - but hasn't for quite some time from what I can tell (I checked on the last stable release, plus any maps using Tech terrain (eg. business as usual) place dirt under the Ore. (Intentional that it can't be placed on Tech?).

Thus partly fixes #1234   (ooh, neat bug number ;P  ).

(I also added Ice and Stone as valid Terrain Ore Terrain types, but - as with Tech - placing ore on these does not actually work in the map editor.)

### DIFF
--- a/mods/hv/rules/world.yaml
+++ b/mods/hv/rules/world.yaml
@@ -302,12 +302,12 @@ World:
 			gold:
 				ResourceIndex: 1
 				TerrainType: Ore
-				AllowedTerrainTypes: Clear, Crater, Grass, Grass Pit, Road, Mountain, Snow, Red Snow, Sand, Plant, Tech, Dirt
+				AllowedTerrainTypes: Clear, Crater, Grass, Grass Pit, Road, Mountain, Ice, Snow, Red Snow, Sand, Black Sand, Plant, Stone, Tech, Dirt
 				MaxDensity: 3
 			iron:
 				ResourceIndex: 2
 				TerrainType: Ore
-				AllowedTerrainTypes: Clear, Crater, Grass, Grass Pit, Road, Mountain, Snow, Red Snow, Sand, Plant, Tech, Dirt
+				AllowedTerrainTypes: Clear, Crater, Grass, Grass Pit, Road, Mountain, Ice, Snow, Red Snow, Sand, Black Sand, Plant, Stone, Tech, Dirt
 				MaxDensity: 3
 			bridge-shadow-left:
 				ResourceIndex: 3
--- a/mods/hv/rules/buildings.yaml
+++ b/mods/hv/rules/buildings.yaml
@@ -619,7 +619,7 @@ MINER2:
 		Footprint: ++ ++ xx
 		Dimensions: 2,3
 		RequiresBaseProvider: false
-		TerrainTypes: Clear, Road, Crater, Grass, Grass Pit, Ore, Mountain, Rock, Ice, Snow, Sand, Black Sand, Stone, Tech, Dirt, Blocked
+		TerrainTypes: Clear, Road, Crater, Grass, Grass Pit, Ore, Mountain, Rock, Ice, Snow, Red Snow, Sand, Black Sand, Stone, Tech, Dirt, Blocked
 		AllowPlacementOnResources: true
 	RevealsShroud:
 		Range: 5c0


mods/hv/rules/world.yaml
```diff
@@ -302,12 +302,12 @@ World:
...
 				ResourceIndex: 1
 				TerrainType: Ore
-				AllowedTerrainTypes: Clear, Crater, Grass, Grass Pit, Road, Mountain, Snow, Red Snow, Sand, Plant, Tech, Dirt
+				AllowedTerrainTypes: Clear, Crater, Grass, Grass Pit, Road, Mountain, Ice, Snow, Red Snow, Sand, Black Sand, Plant, Stone, Tech, Dirt
 				MaxDensity: 3
 			iron:
 				ResourceIndex: 2
 				TerrainType: Ore
-				AllowedTerrainTypes: Clear, Crater, Grass, Grass Pit, Road, Mountain, Snow, Red Snow, Sand, Plant, Tech, Dirt
+				AllowedTerrainTypes: Clear, Crater, Grass, Grass Pit, Road, Mountain, Ice, Snow, Red Snow, Sand, Black Sand, Plant, Stone, Tech, Dirt
```
OcelotWalrus, line 310: I think Ice and Stone are intended not to have ores on.

OcelotWalrus:
> (I also added Ice and Stone as valid Terrain Ore Terrain types, but - as with Tech - placing ore on these does not actually work in the map editor.)

You have a
```
                        gold:
				ResourceIndex: 1
				TerrainType: Ore
				AllowedTerrainTypes: Clear, Crater, Grass, Grass Pit, Road, Mountain, Snow, Red Snow, Sand, Plant, Tech, Dirt
				AllowedTerrainTypes: Clear, Crater, Grass, Grass Pit, Road, Mountain, Ice, Snow, Red Snow, Sand, Black Sand, Plant, Stone, Tech, Dirt
				MaxDensity: 3
			iron:
				ResourceIndex: 2
				TerrainType: Ore
				AllowedTerrainTypes: Clear, Crater, Grass, Grass Pit, Road, Mountain, Snow, Red Snow, Sand, Plant, Tech, Dirt
				AllowedTerrainTypes: Clear, Crater, Grass, Grass Pit, Road, Mountain, Ice, Snow, Red Snow, Sand, Black Sand, Plant, Stone, Tech, Dirt
				
```

in the `editor.yaml` file in plus of the one in `world.yaml`, that's why you can't place ore on these terrain types in the editor. You have to edit both files so they're the same for that trait.

SeanScherer:
Thanks Cromha!

I was thinking it might be some kind of editor setting, but I missed that editor.yaml ...

I get your point that Ore on Ice is usually not something that "works" - but for Antarctica style maps, (or frozen urban areas), <strike>I'd like to have the option (guess I could just modify the file for my editor if need be though :P).</strike> I think its nice to have the option.

I don't see why no Ore on Stone, tbh? (I think there'd be a better argument to be had for not having it on Tech - which is maybe why it was disabled - editor side - this whole time?)

OcelotWalrus:
To be honest I'm for allowing ore on any terrain type, it's up to the map maker to decide if it either fits/looks good or not.

mods/hv/rules/world.yaml
Mailaender: Maybe I had that in mind when I set up the tile set, but I did it inconsistently because mountain is actually natural stone and probably just forgot to update this long list. Ice may be debatable because it probably also covers the water transitioning tiles, but I take this as is for now.

PR #1322, opened 2025-07-13
Drop Newtonsoft.Json dependency

Mailaender:
use https://learn.microsoft.com/en-us/dotnet/api/system.text.json?view=net-8.0 instead.


Mailaender:
This depends on https://github.com/OpenHV/OpenHV/pull/1343.

Issue #1309, opened 2025-07-09
Balance Teleporting Power from Televator


WilliamWeberBerrutti:
The idea is basically to only allow teleporting in and out when near a televator.

Maybe create a teleporting region around them, working more less like the Build Area, but specific for Teleporting usage.

For instance, if a player wants to teleport units to the enemy's base, they'll need to proxy a televator first and then teleport to this proxy's teleporting region.

PR #1333, opened 2025-07-16
Adding distance constraints for teleport support power

ivanpmartell:
Addressing #1309 

This is a first draft to being able to only teleport to cells near a televator. Please suggest circle colors or any other improvements required to merge.

### DIFF
--- a/OpenRA.Mods.HV/Traits/SupportPowers/TeleportPower.cs
+++ b/OpenRA.Mods.HV/Traits/SupportPowers/TeleportPower.cs
@@ -61,7 +61,7 @@ namespace OpenRA.Mods.HV.Traits
 		public readonly WDist RangeFromTelevators = WDist.FromCells(10);
 
 		[Desc("Range circle color.")]
-		public readonly Color CircleColor = Color.FromArgb(128, Color.Red);
+		public readonly Color CircleColor = Color.FromArgb(128, Color.MediumSlateBlue);
 
 		[Desc("Range circle width in pixel.")]
 		public readonly float CircleWidth = 1;


OpenRA.Mods.HV/Traits/SupportPowers/TeleportPower.cs
```diff
@@ -56,6 +57,23 @@ sealed class TeleportPowerInfo : SupportPowerInfo
 		[Desc("Cursor to display when the targeted area is blocked.")]
 		public readonly string TargetBlockedCursor = "move-blocked";
 
+		[Desc("Maximum range from televator to be able to teleport.")]
+		public readonly int RangeFromTelevatorsSquared = 100;
```
Mailaender, line 61:
```suggestion
		public readonly WDist RangeFromTelevators = 100;
```
to not leak an implementation detail into game rules.

OpenRA.Mods.HV/Traits/SupportPowers/TeleportPower.cs
```diff
@@ -56,6 +57,23 @@ sealed class TeleportPowerInfo : SupportPowerInfo
...
+
+		public readonly int CellRangeFromTelevators = 10;
+
+		[Desc("Range circle color.")]
+		public readonly Color CircleColor = Color.FromArgb(128, Color.Red);
```
Mailaender, line 66:
```suggestion
		public readonly Color CircleColor = Color.FromArgb(128, Color.MediumSlateBlue);
```
maybe as the nuke is already red.

OpenRA.Mods.HV/Traits/SupportPowers/TeleportPower.cs
```diff
@@ -56,6 +57,23 @@ sealed class TeleportPowerInfo : SupportPowerInfo
 		[Desc("Cursor to display when the targeted area is blocked.")]
 		public readonly string TargetBlockedCursor = "move-blocked";
 
+		[Desc("Maximum range from televator to be able to teleport.")]
+		public readonly WDist RangeFromTelevators = 100;
```
OcelotWalrus, line 61: Maybe it should be a percentage of the map's size? It'd seem appropriate to me. It could also be interesting to add a "televator weight" attribute to all ground and naval units, and make so that depending on the weight of all units that are being teleported, the range increases or decreases.

WilliamWeberBerrutti: > Maybe it should be a percentage of the map's size?

That's an interesting idea, but I think, gameplay-wise, could be better a constant value. Being a percentage of the map's size would be difficult for the opponent player to predict the presence of enemy Televators.

PR #1352, opened 2025-07-23
Make Miner deploy on ore rallypoints

ivanpmartell:
Allows miners to AttackMove to rallypoint cells without ore and deploy on rallypoint cells with ore. 
It requires creationRallypoint from Mobile.cs to be made public (Although this field can also be obtained through reflection if necessary).
A change on the LeaveProductionActivity class (in Mobile.cs) is also necessary, as without that change the miner will go through all the AttackMove rallypoints before going back to the deploy rallypoints.

These required changes to the engine are documented in OpenRA/OpenRA#21986

### DIFF
--- a/OpenRA.Mods.HV/Traits/Miner.cs
+++ b/OpenRA.Mods.HV/Traits/Miner.cs
@@ -38,7 +38,7 @@ namespace OpenRA.Mods.HV.Traits
 		public override object Create(ActorInitializer init) { return new Miner(this, init.Self); }
 	}
 
-	public class Miner : IIssueOrder, IResolveOrder, IOrderVoice
+	public class Miner : IIssueOrder, IResolveOrder, IOrderVoice, INotifyCreated
 	{
 		public readonly MinerInfo Info;
 		readonly IResourceLayer resourceLayer;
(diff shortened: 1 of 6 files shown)


ivanpmartell:
Rebased, and no more engine changes needed

ivanpmartell:
This does ActivityQueue hacking and is bad practice. Will shelve and try a better alternative.

OpenRA.Mods.HV/Traits/Miner.cs
```diff
@@ -49,6 +53,43 @@ public Miner(MinerInfo info, Actor self)
...
 		}
 
+		void INotifyAddedToWorld.AddedToWorld(Actor self)
+		{
+			var activity = self.CurrentActivity;
+			if (activity is not LeaveProductionActivity)
+				return;
+
+			var mobile = self.Trait<Mobile>();
+			var fieldInfo = mobile.GetType().GetField("creationRallypoint", BindingFlags.NonPublic | BindingFlags.Instance);
```
Mailaender, line 63: Don't ever use reflection needlessly!

PR #1357, opened 2025-07-25
Fix queued miner deploy command not working in the middle of queued activities

### LINKED ISSUE #1356: Miner will not deploy on queued activities
Commit: c3ec817ebf6d94e939db3f371ad0c5c7b4610796

Description:
When shift-queuing commands, if a DeployMiner activity is issued but another command is queued afterwards. e.g. move somewhere else, the miner will not skip deploying. Currently, deployment only works if the DeployMiner activity is the last one in the queue.

Reproduction:

1. Select miner
2. Right click on ore patch
3. Shift-click to move somewhere else far from the ore
4. Miner will not deploy on the ore patch when it reaches that location, moving to the next waypoint.

Expected Behaviour:

The miner should deploy on available ore patch instead of going to the next location.

ivanpmartell:
Fixes #1356

This bug occurs because the Transform activity after the DeployMiner one gets queued at the end of all activities. Thus, it will run Transform at the last waypoint (not in the ore), and fail.


OpenRA.Mods.HV/Properties/launchSettings.json
```diff
@@ -0,0 +1,13 @@
+{
+  "profiles": {
+    "OpenRA.Mods.HV": {
+      "commandName": "Project"
+    },
+    "OpenHV": {
+      "commandName": "Executable",
+      "executablePath": "C:\\Users\\IvanP\\Downloads\\OpenHV-BugFixes\\engine\\bin\\OpenRA.exe",
```
Mailaender, line 8: Please don't hardcode paths that only work on your computer.

OpenRA.Mods.HV/OpenRA.Mods.HV.csproj.user
Mailaender, line 1: This seems to be user specific files. Please put them into `.gitignore`.

ivanpmartell:
Rebased and removed the launch and user files. Also added them to .gitignore

PR #1363, opened 2025-07-25
Make Artillery display an attack range indication when selected

WilliamWeberBerrutti:


Mailaender:
This looks like a workaround for a missing art problem: https://github.com/OpenHV/OpenHV/blob/f872b32f741b550052e48ac267b9d381c86abcef/mods/hv/cursor.yaml#L72-L84

Mailaender:
Those siege tanks get deployed into stationary forms. Artillery is mobile just like the rocket launcher artillery tank. I added https://github.com/OpenHV/OpenHV/commit/db4294740f3b72c52bb4dc8c706c7b4481f56db7 to solve this differently.

PR #1379, opened 2025-08-07
Draft: Prevent all projectiles from being blocked by walls

WilliamWeberBerrutti:


mods/hv/weapons/firearms.yaml
```diff
@@ -18,7 +18,7 @@
 
 ^BlueTracers:
 	Projectile: InstantHitWithTracers
-		Blockable: true
+		Blockable: false
```
Mailaender, line 21: Not a turret weapon.

Mailaender:
But you said you want to prevent exactly that:
> So situations like turrets behind allied walls do not harm them.

WilliamWeberBerrutti:
> I don't think bullets should penetrate walls that are higher than units except for mortars/artillery shells that clearly have a ballistic curve.

Indeed, that's exactly what I thought before, but what can we do to block electric shots, since they behave like bullets? I noticed I can't use `Blockable: true` on those, so, to make matters consistant, I decided to let all projectiles pass thru walls, but I'm open for suggestions.

WilliamWeberBerrutti:
And also regarding turrets placed near walls, what can be done for them? I oftentimes saw them destroying their own walls and not hitting their targets, which is detrimental to player gameplay, in my opinion.

WilliamWeberBerrutti:
> Fix what you wanted to do from the beginning which is turrets hitting walls. There are meant to be used together. If you want walls that can pass through bullets do barb wire or laser barriers. It does not make sense for large concrete walls.

Agreed! This idea makes more sense! I think all comments were resolved!

mods/hv/weapons/firearms.yaml
```diff
@@ -314,6 +318,8 @@ BunkerFlamethrower:
 	Inherits: Flamethrower
 	ReloadDelay: 25
 	Range: 5c0
+	Projectile: Bullet
+		Blockable: false
```
Mailaender, line 322: and boats?

mods/hv/weapons/missiles.yaml
```diff
@@ -137,6 +137,8 @@ RapidPodRocket:
 	Inherits: LightPodRocket
 	ReloadDelay: 23
 	Range: 8c0
+	Projectile: Missile
+		Blockable: false
```
Mailaender, line 141: Rocketeers are infantry which in my mind are smaller than walls.

mods/hv/weapons/firearms.yaml
```diff
@@ -224,6 +226,8 @@ SniperRifle:
 BunkerSniperRifle:
 	Inherits: SniperRifle
 	ReloadDelay: 25
+	Projectile: InstantHit
+		Blockable: false
```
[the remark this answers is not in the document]
WilliamWeberBerrutti: So projectiles from all turrets except bunkers should pass thru walls? I understood bunker ones should too, but I can remove that no problem!

PR #1384, opened 2025-08-08
Make it possible to speed up production of main buildings tab

WilliamWeberBerrutti:


WilliamWeberBerrutti:
> That does not make sense. There can only be one `BASE`.

If you capture enemy bases, you'd have more than 1. This situation is pretty rare, but can happen.

PR #1401, opened 2025-08-29
Fix Footprint for Ore Smelters / Purifiers and Outposts

WilliamWeberBerrutti:


Mailaender:
Can you explain why this change is fixing anything?

WilliamWeberBerrutti:
For instance, units may be hidden behind the buildings, like this (there's 1 Missile Tank behind each of them):

Screenshot_20250829_153756

WilliamWeberBerrutti:
> Well, not sure if you know, but there's also `+` footprint, which allows units to pass through, and the tile won't be blocked for actors. Maybe that would have been better?

Yes, I think it should be better! The documentation is lacking on that part: https://openhv.readthedocs.io/en/latest/release/traits/#building

PR #1412, opened 2025-11-02
Fix: Repair tank auto-targets allies only

DealsBeam:
The REPAIRTANK was incorrectly configured to auto-target any damaged vehicle, including enemies. This change restricts the auto-targeting to allied units only by adding `TargetRelationships: Ally` to the `AutoTargetPriority@Default` trait.


Mailaender:
Please don't submit unchecked AI submissions:

> Error: hv|rules/vehicles.yaml:370 refers to a trait field `TargetRelationships` that does not exist on `AutoTargetPriority`.

I am also pretty sure we don't have a bug where the repair tank actually repairs enemies.

PR #1429, opened 2026-07-26
Add Tracers to Sniper so it's easier to identify where the shots are coming from

WilliamWeberBerrutti:


Mailaender (approved):
Looks very cyberpunk :sunglasses:

PR #1431, opened 2026-08-24
Lower the volume of aircraft chaingun sound

WilliamWeberBerrutti:
Since it's too loud, specially when having groups of Attack Helicopters or Gunships.

The way it was reduced was using Tenacity with `Loudness Normalization` effect, using these settings:

Screenshot_20260824_175422


Dzierzan:
And how does waveform look like after changes? I compared it to other sounds, it's not really any louder than other sounds.

WilliamWeberBerrutti:
> And how does waveform look like after changes?

It looks like this:

image

> I compared it to other sounds, it's not really any louder than other sounds.

It's loud when you have like 5 or more helicopters / planes attacking simultaneously. Rampoina and me have to lower the sound volume to be tolerable, but that makes all the other sounds less noticeable too.

I'm a complete noob regarding sound edits so I'm not sure if it's an adequate change haha

